PR #4, opened 2017-05-09
Patch 1

qubusp:
More modern repo for docker, where it's necessary to use only docker-ce


geerlingguy:
This was fixed in #6 (Issue #5), and should work with `docker-ce` or `docker-ee` on CentOS, Ubuntu, and Debian (at least... haven't tried on other distros yet).

Issue #8, opened 2017-05-21
Add `docker` python package


triplepoint:
It's plausible someone is installing this role because they're planning to execute an Ansible `docker_container` task, however this task depends on the python `docker` package to function.

It'd be great if this role added this package directly, to avoid having to check for it in dependent roles.

geerlingguy:
Yeah... and I just ran into:

```
TASK [Build Docker image from Dockerfiles.] ************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to import docker-py - No module named requests.exceptions. Try `pip install docker-py`"}
```

That happened during a [test run](https://travis-ci.org/geerlingguy/ansible-for-devops/jobs/236586398) of the code backing an example in Ansible for DevOps. It would definitely be convenient to have this role be a one-stop shop for installing `docker-py`...

geerlingguy:
...but as I mentioned in [this PR comment](https://github.com/geerlingguy/ansible-role-docker/pull/9#issuecomment-302974457), it's probably more architecturally sound to split that off into a `pip` role. Therefore, I present to you: https://github.com/geerlingguy/ansible-role-pip

To get `docker-py`, you can add the role `geerlingguy.pip`, and in your playbook variables, set:

```yaml
pip_install_packages:
  - name: docker-py
    version: "1.10.6"
```

I'm going to add a note to that effect in this role's documentation/README before closing out this issue, though.

triplepoint:
Right on, I had a similar role being copypasta'ed between a few different projects; probably worth scraping that out and replacing it with a published dependency.  Thanks for taking care of the documentation, I can call this one resolved.

geerlingguy:
Perfect! I updated the documentation in the commit above.

triplepoint:
Did you specifically mean `docker-py` and not `docker`?  It looks like the `docker-py` project renamed their package near the end of last year, with their 2.0 release.
https://docker-py.readthedocs.io/en/stable/change-log.html#

geerlingguy:
Gah! The level of renames that occur in the Docker ecosystem is annoying.

I'll update the docs and usage...

PR #9, opened 2017-05-21
Add installation of the docker python package

triplepoint:
Addresses issue #8

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -14,5 +14,8 @@
     state: started
     enabled: yes
 
+- name: Ensure docker Python package is available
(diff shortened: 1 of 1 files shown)


triplepoint:
Ungh, nevermind, we can't assume pip is in place either, and this seems like a bad place to go installing something like a python package manager.

geerlingguy:
Heh, yeah. I think either doing that in the upstream playbook, or maybe better using a python pip role (if one doesn't exist, I could whip one up) that installs pip then uses Ansible's pup module to install a variable list of packages.

Issue #7, opened 2017-05-21
Role won't succeed on ubuntu 14.04


triplepoint:
Looks like the older version of Python and it's lack of SNI support strikes again.  

This role won't succeed against Ubuntu 14.04, due to its python `2.7.5` version and the lack of support for SNI.

I guess download.docker.com didn't used to require SNI?  This sort of thing sometimes happens when services change CDN providers.

Might consider adding ubuntu 14.04 to travis, to catch situations like this.

Log attached:
```
TASK [geerlingguy.docker : Add Docker apt key.] ********************************
changed: [ubuntu_16_04]
fatal: [ubuntu_14_04]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to validate the SSL certificate for download.docker.com:443. Make sure your managed systems have a valid CA certificate installed. If the website serving the url uses SNI you need python >= 2.7.9 on your managed machine or you can install the `urllib3`, `pyopenssl`, `ndg-httpsclient`, and `pyasn1` python modules to perform SNI verification in python >= 2.6. You can use validate_certs=False if you do not need to confirm the servers identity but this is unsafe and not recommended. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible. The exception msg was: [Errno 1] _ssl.c:510: error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure."}```

**Comment by kostyrev** [comment]
BTW trusty is listed in meta/main.yml as supported OS.

**Comment by geerlingguy** [comment]
Well, it _was_ working when I originally set it up :)

**Comment by geerlingguy** [comment]
I'll need to add a Travis CI test on `ubuntu1404` and see if I can't get it happy.

**Comment by kostyrev** [comment]
We are just stating the fact and not blaming you.
Thanks for the role and its support.

**Comment by triplepoint** [comment]
PR #10 did end up adding a dependency on cURL that didn't previously exist, so for the vanishingly small case of "has old python version and no cURL", this still fails, but that's probably not a serious issue.  I ran into it deploying to a test docker image for a `molecule` test of another role, but I can't imagine a real world target host not having cURL. I'm ok with this being closed.

**Comment by geerlingguy** [comment]
@triplepoint - Yeah, and worst case I'd recommend someone slip in a:

```
- package: name=curl state=present
```

[...]

PR #10, opened 2017-05-23
Fix adding docker apt key on Ubuntu 14.04

kostyrev:
Dirty fix for #7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ services: docker
 env:
   - distro: centos7
   - distro: ubuntu1604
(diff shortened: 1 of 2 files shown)


geerlingguy:
@kostyrev thanks for adding this—can you also add to the `.travis.yml` configuration the line:

```yaml
  - distro: ubuntu1404
```

(After the line for 1604). That way tests will also be run against 14.04. Thanks!

geerlingguy:
Works for me! I'll change a tiny bit for better logging.

PR #13, opened 2017-07-18
Fix adding docker GPG key & repo on Centos 7.1

kakawait:
Same problem as #7

### DIFF
--- a/tasks/setup-RedHat.yml
+++ b/tasks/setup-RedHat.yml
@@ -12,6 +12,28 @@
   rpm_key:
     key: https://download.docker.com/linux/centos/gpg
     state: present
+  register: add_repository_key
+  ignore_errors: true
+
+- name: Ensure curl is present (on older systems without SNI).
+  package: name=curl state=present
+  when: add_repository_key|failed
+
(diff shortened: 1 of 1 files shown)


kakawait:
@geerlingguy Unfortunately you don't have `distro: centos71`

geerlingguy:
I'm only testing against whatever's the current version of CentOS 7—does the role not work against 7.3? Also, since I don't have a use case, what would be a condition where CentOS 7.1 would need to be locked in/continued to be used (vs. upgrading to or installing by default 7.3)?

kakawait:
I didn't test myself but I really think role works great on Centos 7.3. 

My use case is close to stupid use case but due to _certification_ stuffs, on my company we're currently blocked to use Centos 7.1.1503. Just that... That why I proposed that PR (tested in my case) but I could understand that a bit too edge case and not receivable.

geerlingguy:
Heh, I totally understand. I'd rather not merge it since I support and recommend the latest release (and usually n-1, which would be 7.2) for the distributions I support. I'd recommend for now you use a fork of this role for the company in question, and then switch back to this role or another once you can upgrade versions.

PR #15, opened 2017-07-20
Add var to customize docker pkg state

### LINKED ISSUE #14: Provide a way to upgrade Docker package
What do you think about offering the possibility to upgrade Docker version by simply not hard-code `state=present` 

Example from other role (random pick) https://github.com/angstwad/docker.ubuntu/blob/master/tasks/main.yml#L97

kakawait:
By default var will be `present` to do not break legacy.

Fixes #14


geerlingguy:
Adding docs in a follow-up.

PR #16, opened 2017-07-20
Open up support for armhf architecture

digmore:
Nice module and thanks for making it available!

Could you please consider this PR to add armhf package support to the default configuration? I know I can override the repository URL in my playbook, but it would be nice for it to "just work" for both amd64 and armhf architectures. By removing this bit apt should fall back on the value in the apt configuration.

There are also s390x packages for Xenial:
https://download.docker.com/linux/ubuntu/dists/xenial/stable/
I presume this would add support for that architecture too, though do not have access to such a system to test.


geerlingguy:
Passes tests, so definitely! Thanks for the PR.

PR #17, opened 2017-08-04
Eliminated the need for two OS specific tasks

PhilThurston:
First of all, great role! This role will help a great deal and this is the first of several pull requests where we have changed this to suit our needs and think others would benefit as well.

There are two changes that have been made:

### 1. Eliminated the need for two OS specific tasks
I like to keep all of our plays built with the "K.I.S.S" method. "Keeping it sweet and simple." That means I prefer to have as few "skipped" tasks as I possibly can in a play. Including OS specific tasks can be done with a single task this eliminates at least one task from your plan and giving an overall cleaner output. Very minor I know.

### 2. Added ability to completely skip the role.
There may be times where a single host in a group shouldn't have docker installed and you want to ensure that it is never mistakenly installed. Simply adding `docker_install: false` and `docker_install_compose: false` to the host_vars for the remote. This will ensure that anytime the role is ran against the hosts group it will simply skip the installation of the tasks.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,9 +1,6 @@
 ---
-- include: setup-RedHat.yml
-  when: ansible_os_family == 'RedHat'
-
-- include: setup-Debian.yml
-  when: ansible_os_family == 'Debian'
+- name: OS specific setup
+  include: include: "setup-{{ ansible_os_family }}.yml"
 
 - name: Install Docker.
   package: name={{ docker_package }} state={{ docker_package_state }}
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
 - name: OS specific setup
-  include: include: "setup-{{ ansible_os_family }}.yml"
+  include: "setup-{{ ansible_os_family }}.yml"
   when: docker_install
 
 - name: Install Docker.
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,15 +1,18 @@
 ---
 - name: OS specific setup
   include: include: "setup-{{ ansible_os_family }}.yml"
+  when: docker_install
 
 - name: Install Docker.
   package: name={{ docker_package }} state={{ docker_package_state }}
+  when: docker_install
 
 - name: Ensure Docker is started and enabled at boot.
   service:
     name: docker
     state: started
     enabled: yes
+  when: docker_install
 
 - include: docker-compose.yml
   when: docker_install_compose
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,5 +1,6 @@
 ---
 # Edition can be one of: 'ce' (Community Edition) or 'ee' (Enterprise Edition).
+docker_install: true
 docker_edition: 'ce'
 docker_package: "docker-{{ docker_edition }}"
 docker_package_state: present


tasks/main.yml
```diff
@@ -1,15 +1,18 @@
 ---
 - name: OS specific setup
   include: include: "setup-{{ ansible_os_family }}.yml"
+  when: docker_install
```
it-praktyk: I don't understand a reason to add skip_install to the role what is intended only to install

PhilThurston: The reason for this change is really for safety and to guard against user error. If you're so inclined you can explicitly state if a host should or should not have docker installed on the system. This prevents an admin from accidentally installing docker on a host that shouldn't have it installed in the first place. So take the following as an example.
```
[webhost]
host1
host2
host3 <== this host shouldn't ever have docker installed.
host4
```

Now, lets say that an admin some time down the road forgets taht `host3` shouldn't have docker installed and puts the webhost group in the playbook because lets say the docker repo needs to be updated across all docker hosts. The role will modify docker on all those hosts but will also install docker on `host3`. This change simply gives the option to have that extra safety there so you can make sure some hosts never get it installed regardless of user error.

it-praktyk:
I don't agree with your way of thinking. For scoping of execution of tasks are other mechanisms than a variable in the role.

If you need to be sure that something will not be executed on the particular host you should:
- refactor your inventory,
- invoke rule execution using the [include_role](https://docs.ansible.com/ansible/latest/include_role_module.html) module with the `when` statement and e.g. hosts variables.

BTW, I think that:
- your idea in https://github.com/geerlingguy/ansible-role-docker/pull/17/commits/98ae2332691836d7d327615e270772cb620b7a17 is good
- implementation of https://github.com/geerlingguy/ansible-role-docker/pull/17/commits/98ae2332691836d7d327615e270772cb620b7a17 could be more elegant with a [block](https://docs.ansible.com/ansible/latest/playbooks_blocks.html)

PhilThurston:
Ansible lists in it's best practices to [keep it simple](http://docs.ansible.com/ansible/latest/playbooks_best_practices.html#keep-it-simple) there is no need to over complicate things. Refactoring an inventory just for something as simple as a docker install is overkill especially for use-cases where there would be 1-2 exceptions to the rule. Additionally the `include_roles` module just adds unneeded complexity to the role. Why not just have the variable? It is completely optional to use and doesn't affect the role in any way unless you decide to use it. A simple solution to a simple problem.

Additionally why would we want to complicate 98ae2332691836d7d327615e270772cb620b7a17 ? The solution in that commit is very simple and gets the job done without a bunch of additions or unneeded complexity. You instantly know what is happening at a glance. There is no need to over-complicate things with using a block. 

To quote [Ansible for Devops](https://www.ansiblefordevops.com/)
> It is perfectly acceptable to try a different approach... being consistent is more important than following a particular set of rules, especially if the ruleset isn't universally agreed upon.

I, myself and our team as a whole are down for trying new things as long as it is a methodology that doesn't complicate things. We are just being consistent in what we use with our other plays/roles and we love these solutions as a team and use it in our own flavor of our existing docker role. We're always willing to change though. 

If you have a **simpler solution** to what I've proposed in this pull request then I would absolutely love to see it. Simply give an example task in the comments. We run with the credo of simplicity and consistency in our team and we use these methods across dozens of plays and roles between a full team of people. After lots of trial and error this seems to be a solution we, as a whole find is most efficient.

tasks/main.yml
```diff
@@ -1,18 +1,18 @@
 ---
-- include: setup-RedHat.yml
-  when: ansible_os_family == 'RedHat'
-
-- include: setup-Debian.yml
-  when: ansible_os_family == 'Debian'
+- name: OS specific setup
+  include: "setup-{{ ansible_os_family }}.yml"
+  when: docker_install
```
geerlingguy: If you'd like to skip installing docker... then just skip running this role?

I'm not sure what the use case is for running this role but not actually using it to do anything.

Usually, if I have a playbook that needs to do something conditionally (e.g. add a role), I'd add the conditional in the playbook itself, e.g.:

```yaml
- hosts: all
  roles:
    - role: geerlingguy.docker
      when: docker_install
```

(You could use any conditional you'd like there.)

tasks/main.yml
```diff
@@ -1,18 +1,18 @@
 ---
-- include: setup-RedHat.yml
-  when: ansible_os_family == 'RedHat'
-
-- include: setup-Debian.yml
-  when: ansible_os_family == 'Debian'
+- name: OS specific setup
+  include: "setup-{{ ansible_os_family }}.yml"
```
geerlingguy: I'm okay with the setup change (though to follow my role syntax style, please format it like this:

```yaml
# Setup/install tasks.
- include: "setup-{{ ansible_os_family }}.yml"
```

...but please remove the conditional (see reasoning elsewhere in this review).

PR #18, opened 2017-08-06
Adds an initial Docker daemon configuration for DNS resolvers

jbenden:
Simple PR to add a basic Docker daemon configuration, which only adds DNS resolvers at this point.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -8,6 +8,15 @@
 - name: Install Docker.
   package: name={{ docker_package }} state={{ docker_package_state }}
 
+- name: Ensure Docker configuration is present.
+  template:
+    src: etc/docker/daemon.json.j2
+    dest: /etc/docker/daemon.json
+    owner: root
+    group: docker
+    mode: 0640
+  notify: restart docker
+
 - name: Ensure Docker is started and enabled at boot.
   service:
     name: docker
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -8,6 +8,14 @@
 - name: Install Docker.
   package: name={{ docker_package }} state={{ docker_package_state }}
 
+- name: Ensure Docker directory exists.
+  file:
+    path: /etc/docker
+    state: directory
+    owner: root
+    group: root
+    mode: 0700
+
 - name: Ensure Docker configuration is present.
   template:
     src: etc/docker/daemon.json.j2
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -9,6 +9,11 @@ docker_install_compose: true
 docker_compose_version: "1.13.0"
 docker_compose_path: /usr/local/bin/docker-compose
 
+# Docker Daemon Configuration Options.
+docker_daemon_dns_resolvers:
+  - "8.8.8.8"
+  - "8.8.4.4"
+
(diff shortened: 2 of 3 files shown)


PhilThurston:
Awesome, @jbenden. You almost have it. We have this right before our template file. You need to manually make the directory. Docker makes this directory after it is started for the first time but if you are doing key things like setting `graph` which changes the core directory of docker then you want this json file in place before you run it for the first time.  If you add the below snippet before the template. Once you've done this it will pass pipeline.

```
- name: Make sure Docker directory exists.
  file:
    path: /etc/docker
    state: directory
    mode: 0700
```

jbenden:
Thanks @PhilThurston for the tip!

PhilThurston:
No problem! One thought I'd like to add just as an opinion. There are so many options available for the config file that making a variable driven template file would be a daunting task. It may be just easier to have a, maybe, blank template file that users can edit on their own and then can customize via host/group specific templates. Something like the following could work if you put the customized file in that directory

```
- name: Ensure Docker configuration is present.
  template:
    src: "{{ inventory_hostname }}/daemon.json.j2"
    dest: /etc/docker/daemon.json
    owner: root
    group: docker
    mode: 0640
  notify: restart docker
```

A blank daemon would also work just great and let the users edit it but that would leave the issue of customization between different hosts. Another thing you could do but it feels very "hacky" is to have a jinja conditional for the file to use:

```
- name: Ensure Docker configuration is present.
  template:
    src: "{% if daemon.json.stat.exists %}{{ inventory_hostname }}/daemon.json.j2{% else %}daemon.json.j2{% endif %}"
    dest: /etc/docker/daemon.json
    owner: root
    group: docker
    mode: 0640
  notify: restart docker
```
This would be done with a local check for the file, but I haven't tested this. My favorite option though would probably be to just make a variable array that can be added into the file and have the template file do a loop through the array. Sort of like you have it now but instead of it setting DNS in every file; instead each item the user has defined. You'd do this with, I think, nested loops.

```
docker_daemon_config:
  - dns: [8.8.8.8,8.8.4.4]
  - iptables: [false]
  - graph: [/home/docker]
```
Again this is untested, I don't even know if my syntax is correct just throwing ideas out there because I see with the current implementation that this config file and default vars can snowball quickly. :smile

tasks/main.yml
```diff
@@ -8,6 +8,23 @@
...
+  file:
+    path: /etc/docker
+    state: directory
+    owner: root
+    group: root
+    mode: 0700
+
+- name: Ensure Docker configuration is present.
+  template:
+    src: etc/docker/daemon.json.j2
```
geerlingguy: If I'm going to include any template with this role, I'd like to do it in as flexible a manner as possible. I do like the idea of having at least a few sane defaults (I have often had to switch on Google DNS for container resolution to work...), but it should be easy for someone to completely override the template, and/or possibly blocks inside the template.

For this example taken to the extreme, check out my Nginx role: https://github.com/geerlingguy/ansible-role-nginx#overriding-configuration-templates

For a simpler example, check out my Varnish role: https://github.com/geerlingguy/ansible-role-varnish#role-variables (the `varnish_default_vcl_template_path`)

PR #19, opened 2017-08-06
Ubuntu 1604 dmsetup

jbenden:
This patch adds a required package on Ubuntu 16.04, which is required to work-around an upstream bug docker/docker#23347

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -16,6 +16,18 @@
     - apt-transport-https
     - ca-certificates
 
+# Fix for https://github.com/docker/docker/issues/23347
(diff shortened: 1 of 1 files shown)


geerlingguy:
@jbenden - I'm not quite sure what this solves—I have been using this role on a number of 16.04 instances (and it is tested daily against 16.04 here: https://travis-ci.org/geerlingguy/ansible-role-docker/builds), and I haven't run into the problems I'm seeing in that upstream issue.

I'd rather not include fixes that occur in very specialized circumstances... but if you can find a way to reproduce the problem in a reproducible build, I may consider adding a fix. Otherwise, I recommend adding the bugfix to your playbook (or a separate role).

PR #20, opened 2017-08-06
Ensure the updatedb does not include Docker images

jbenden:
This PR ensures that Docker images are not indexed on an Ubuntu server.


geerlingguy:
@jbenden - Is this documented more somewhere? I'm presuming this is for the `locate` command... which isn't necessarily present in all cases. IMO, this looks like an optimization that would be better to have in a different role or the playbook itself. I haven't personally run into any problems with `updatedb`, but what's the reason you would want to make this change? (or more succinctly, what's the problem this is fixing?)

Issue #22, opened 2017-08-16
Recommended packages are not installed for Ubuntu 14.04 (Trusty)


it-praktyk:
The quote from Docker documentation 

> Unless you have a strong reason not to, install the linux-image-extra-* packages, which allow Docker to use the aufs storage drivers.
> 
> ```
> $ sudo apt-get update
> 
> $ sudo apt-get install \
>    linux-image-extra-$(uname -r) \
>   linux-image-extra-virtual
> ```

Source: [Recommended extra packages for Trusty 14.04](https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/#recommended-extra-packages-for-trusty-1404)

Recommended packages are not installed by the current role. Should be?

PhilThurston:
Good catch!

geerlingguy:
I won't necessarily close this issue yet... but I haven't had any issues running Docker on Ubuntu 14.04 with this role, and it seems the daily cron build that runs on 14.04 hasn't had any installation failures either... https://travis-ci.org/geerlingguy/ansible-role-docker/builds/265841483

PR #23, opened 2017-08-17
Added packages for Ubuntu Trusty

PhilThurston:
As mentioned in #22 the packages for docker in Trusty 14.04 are missing from the role. This pull should resolve that issue.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,5 +1,6 @@
 ---
 # Edition can be one of: 'ce' (Community Edition) or 'ee' (Enterprise Edition).
+docker_install: true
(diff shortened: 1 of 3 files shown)


PhilThurston:
This contains the commits from another pull. I thought it would be pulled by now. I'll just leave it.

geerlingguy:
I have a daily build that seems to be working fine on Ubuntu 14.04, even without adding extra packages: https://travis-ci.org/geerlingguy/ansible-role-docker/builds/265841483

I'm not sure if this is required, and it looks like there are additional changes in here that I'd not merge if trying to just add additional packages.

PR #24, opened 2017-08-18
added required packages installation as without those was not able to…

Alex2357:
Added some packages for Debian which are required to create a container. Below is the playbook example which failed without those commands.

---

- hosts: all
  roles:
    - geerlingguy.docker
  tasks:
    - name: create gitolite--docker container
      docker_container: 
        name: mongodbdocker
        image: mongo:3.2
        state: started
        ports:
          - "8082:27017"
        volumes:
          - "/docker/volumes/catalogermongodbdatadir:/data/db"
      become: yes

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -38,3 +38,16 @@
     repo: "{{ docker_apt_repository }}"
     state: present
     update_cache: yes
+
+- name: Install packages required for docker_container see https://docs.ansible.com/ansible/docker_container_module.html#requirements-on-host-that-executes-module
+  apt: name={{item}} state=installed
+  become: yes
+  with_items:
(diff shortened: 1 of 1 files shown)


geerlingguy:
@Alex2357 - From the README (see: https://github.com/geerlingguy/ansible-role-docker#use-with-ansible-and-docker-python-library):

### Use with Ansible (and `docker` Python library)

Many users of this role wish to also use Ansible to then build Docker images and manage Docker containers on the server where Docker is installed. In this case, you can easily add in the docker Python library using the `geerlingguy.pip` role:

```yaml
- hosts: all

  vars:
    pip_install_packages:
      - name: docker

  roles:
    - geerlingguy.pip
    - geerlingguy.docker
```

PR #28, opened 2017-09-24
Comply with Ansible 2.4 new `include` method

pgilad:
Since the variables are computed in parent playbook, I reckon using `import_tasks` (static) is more appropriate than `include_tasks` (dynamic). The ansible docs are very unclear on it:
http://docs.ansible.com/ansible/latest/playbooks_reuse.html#differences-between-static-and-dynamic

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,8 +1,8 @@
 ---
-- include: setup-RedHat.yml
+- import_tasks: setup-RedHat.yml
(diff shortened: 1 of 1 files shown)


geerlingguy:
These need to be `include_tasks` so they work with the dynamic `when` statements (import only works for statically-included files). See: https://github.com/geerlingguy/ansible-role-docker/pull/37

PR #35, opened 2017-10-21
Fix a typo in README.md

daniloradenovic:


geerlingguy:
We would need to make sure this documentation works with both RedHat and Debian derivatives (yum/dnf and apt); one uses the dash, the other the equals sign.

geerlingguy:
Please see above commit ^

PR #37, opened 2017-11-27
'include' for tasks has been deprecated

pad92:
The use of 'include' for tasks has been deprecated. Use 'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,8 +1,8 @@
 ---
-- include: setup-RedHat.yml
+- include_tasks: setup-RedHat.yml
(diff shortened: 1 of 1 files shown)


senorsmile:
We should probably wait to merge this until the "include" directive is fully removed... or at least until 2.3 is no longer supported.  I know some will have to wait to upgrade to 2.4.  And, even though it's deprecated in 2.4, it won't be completely removed for a few more versions.

geerlingguy:
Thanks! Merged and bumped version requirement to Ansible 2.4 due to the change.

PR #38, opened 2017-12-01
Fix apt repo arch issue

senorsmile:
There is an issue (at least on ubuntu 14.04) where if you do not specify the arch (e.g. [arch=amd64]) then it will default to i386.  This obviously fails on an x86_64 machine.  

The official docker install docs for Debian/Ubuntu specify that this must be part of the apt repo config.  

This patch adds a line both for amd64 and armv7l.  There may be more arm types that conform to armhf that I'm not aware of.

Edit: What's strange about this issue is that the role works fine at first.  It's not until trying to do an apt update later that the problem is noticed.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -11,7 +11,8 @@ docker_compose_path: /usr/local/bin/docker-compose
 
 # Used only for Debian/Ubuntu. Switch 'stable' to 'edge' if needed.
 docker_apt_release_channel: stable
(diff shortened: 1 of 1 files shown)


senorsmile:
there are now other fixes in master that override this.  While this was an attempt to auto discover, it's not strictly necessary.  closing.

vpArth:
Link to merged fix: https://github.com/geerlingguy/ansible-role-docker/pull/43/commits/e4a6edb4ef3c1295d87cdfc737a5f07d26c3547b

PR #40, opened 2017-12-14
Update docker compose task to use block statement

JeffreyVdb:
I think these 2 tasks should be grouped in a block, we should only install docker-compose when it's not yet installed.

### DIFF
--- a/tasks/docker-compose.yml
+++ b/tasks/docker-compose.yml
@@ -1,20 +1,23 @@
+# vi: set ft=ansible :
 ---
 - name: Check current docker-compose version.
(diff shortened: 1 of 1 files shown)


tasks/docker-compose.yml
```diff
@@ -1,20 +1,23 @@
+# vi: set ft=ansible :
 ---
 - name: Check current docker-compose version.
-  command: docker-compose --version
+  command: "{{ docker_compose_path }} --version"
```
geerlingguy: Can you break out this fix into a separate PR? This is a bug that should probably be fixed separately.

PR #43, opened 2018-01-19
Debian apt amd64

fanatique:
According to the docs (https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/#install-using-the-repository#set-up-the-repository), Docker needs the CPU architecture to be specified in the apt source definition under Debian/Ubuntu.

Otherwise updating the sources fails with „N: Skipping acquire of configured file 'stable/binary-i386/Packages' as repository 'https://download.docker.com/linux/ubuntu xenial InRelease' doesn't support architecture 'i386‘“

Although the correct package has been installed!

This PR introduces an additional config param for Debian/Ubuntu to setting the CPU architecture to either amd64 or i386.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,9 @@ The `docker_edition` should be either `ce` (Community Edition) or `ee` (Enterpri
 
 Docker Compose installation options.
 
+    docker_apt_arch: amd64
     docker_apt_release_channel: stable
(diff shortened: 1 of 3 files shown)


geerlingguy:
Can you please rebase this to master? Also I think it is missing the documentation on the new variable in the README.

PR #50, opened 2018-02-22
Make ignoring repository key error optional

wedi:
Just ignoring an error regarding the Docker GPG key and downloading it
from "some" web page is not desired in any environment.


geerlingguy:
Makes sense.

PR #51, opened 2018-02-26
Prevent apt arch from autoselect, add variable

vpArth:
There was not a way to partially customize ubuntu preferred arch to install, so it can fail with `No "stable/binary-i386/Packages" in Release file` or user must to redefine full `docker_apt_repository` value.
Variable will fix it without close down of support for other architectures.

[Related commit](https://github.com/geerlingguy/ansible-role-docker/commit/404cfb80d6e0fbcab6f7a04ca6bd56e7819ab535)


vpArth:
@senorsmile, looks pretty similar, your PR should fix my issue too.

geerlingguy:
Closing in favor of #38 — that PR just needs a rebase (or pull out the 2nd commit) and a docs update and it looks good to me.

Issue #36, opened 2017-11-08
Feature request - Set proxy enviroment


jmcarranza:
Hi, it would be very useful to define the proxy variable for docker.

denics:
Hi @matevarga I have the environment set, but the proxy is not working for me.

mxdlx:
Anyone interested in this can add a template and handle it with Ansible following the [official documentation](https://docs.docker.com/config/daemon/systemd/#httphttps-proxy).

Sprinterfreak:
The point where everything breaks in a proxyed environment is due to a bug, called feature not implemented, in `apt_key`. Ansibles apt_key does not pass required additional command-line parameters to use the proxy from environment.

Possible fix is to *NOT* use apt_key's url parameter at all like so:
```
name: Add Docker apt key.
  apt_key:
    data: "{{ lookup('url', 'https://download.docker.com/linux/debian/gpg', split_lines=False) }}"
    id: 9DC858229FC7DD38854AE2D88D81803C0EBFCD88
    state: present
  register: add_repository_key
  ignore_errors: "{{ docker_apt_ignore_key_error }}"
```

PR #55, opened 2018-03-06
Add proxy support using systemd drop-in

riponbanik:
I have looked at another pull request regarding proxy but this I think will be better way to use manage proxy using systemd drop-in


gbonnefille:
I fear you forgot to push the 'tasks' related part of your contribution. Did I missed something?

gbonnefille:
Will fix #66 and #36.

PR #58, opened 2018-04-23
make adding the Docker repo optional

bverschueren:
- organize the repo actions in one block
- make adding the Docker repo optional, which you might not want when using a local Satellite-managed repo

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -5,6 +5,9 @@ docker_package: "docker-{{ docker_edition }}"
 docker_package_state: present
 docker_restart_on_package_change: True
 
+# Add Docker repo
+docker_add_repository: true
+
 # Docker Compose options.
 docker_install_compose: True
 docker_compose_version: "1.21.1"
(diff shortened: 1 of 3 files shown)


geerlingguy:
Before merging, I would need this to apply equally to the Debian setup tasks as well.

geerlingguy:
Looks good to me, can you rebase and I'll merge?

bverschueren:
@geerlingguy: rebased and pushed again (twice; forgot to squash at first). Second ci build failed (temporary issue ?) so I closed/reopened the PR to trigger a new build which passed.

gbonnefille:
I'm really interested in such PR because I really want to deploy distribution-packaged version of Docker.

For example, in RedHat it could be really useful to use their packaged version instead of upstream one (depending on the needs)

Legends0:
@gbonnefille It looks like the PR still runs an uninstall of the distribution's version of docker each time it is runs, which would then lead to an additional install and thus at least two change notices when re-run on system.

PR #68, opened 2018-05-17
Fix version pinning format in README

rnestler:
I had to use `=` to get it to work in Ubuntu. I'm not sure if different distributions may behave differently here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Available variables are listed below, along with default values (see `defaults/m
     docker_package: "docker-{{ docker_edition }}"
     docker_package_state: present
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
Yeah, this seems like that's the correct way when using apt (Debian/Ubuntu/etc.); but the current format is how it should be on Red Hat/CentOS/Fedora.

See: https://unix.stackexchange.com/a/151690/16194

So we might want to make it so it's like:

> (Red Hat/CentOS: `docker-{{ docker_edition }}-<VERSION>`, or Debian/Ubuntu: `docker-{{ docker_edition }}=<VERSION>`)

geerlingguy:
Please see above commit ^

PR #71, opened 2018-05-23
Update user's supplementary group membership, not primary group

### LINKED ISSUE #70: User's primary group shouldn't be updated
When the `docker_users` var is used, the user's primary group is set to `docker`.  This is probably undesirable in most cases, and instead `docker` be added to the list of the user's supplementary groups.

yankcrime:
Fixes #70.


geerlingguy:
Oops, yeah, good catch.

PR #74, opened 2018-07-06
Fix docker_users conditional

larsmaes:
When defining users in the docker_users var, an error occurs about evaluating the docker_users conditional. This change will fix the error


geerlingguy:
`docker_users` should _always_ be defined, as it's a default: https://github.com/geerlingguy/ansible-role-docker/blob/master/defaults/main.yml#L30

Issue #75, opened 2018-07-10
Feature Request - Swarm clustering


colonelpopcorn:
I've had some trouble getting your Kubernetes role to work in a Vagrant environment so I decided to try Docker Swarm and it was a whole lot easier to create a cluster. Yeah, I know it's not as complicated but it would be nice to have some functionality similar to your Kubernetes role in this role. I've got an idea on how to set it up and I'd like to get a pull request in for it. Please let me know if this something you'd be interested in having as a part of this role.

geerlingguy:
Haha, I have stories—_many_ horrible, war stories—from getting Kubernetes running on Vagrant. But here are a couple examples where I do have it working after working through some fun networking things...

  - https://github.com/geerlingguy/ansible-for-devops/tree/master/kubernetes
  - https://github.com/geerlingguy/raspberry-pi-dramble

As mentioned in the PR, I would rather not add swarm support to _this_ role, but rather I think it should be a simple but separate role.

colonelpopcorn:
I published my own role [here](https://github.com/colonelpopcorn/ansible-role-docker-swarm). However, it seems I might have duplicated some other roles. Still, I think it was a good learning experience. Thanks!

PR #76, opened 2018-07-10
Swarm cluster

colonelpopcorn:
This should add the feature proposed in #75. I borrowed heavily from your Kubernetes role to architect it. Please let me know if any changes need to be made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,15 @@ Docker Compose installation options.
 
 A list of system users to be added to the `docker` group (so they can use Docker on the server).
 
+    docker_init_swarm: False
+    docker_swarm_role: 'manager'
+    docker_swarm_advertise_address: ''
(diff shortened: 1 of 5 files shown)


geerlingguy:
@colonelpopcorn - I like the idea, but I don't want to have to test/support Swarm for this role, since (at least IME) most users of this role just want Docker (and many don't even want Compose). I think it would be best to have a separate `namespace.swarm` role (and it would be awesome if you'd create one and put it on Galaxy :) ).

gumbo2k:
If somebody stumbles over this thread looking for swarm support (as I did), there is now a nice role on galaxy named `atosatto.docker-swarm` which does exactly that swarm deployment. 
It can also be reduced to only deploying the docker engine via `skip_swarm: true` and other options.

PR #78, opened 2018-07-22
Fix deprecation warning when installing dependencies; fixes #77

### LINKED ISSUE #77: Invoking "apt" only once while using a loop via squash_actions is deprecated
Using version 2.5.0, I receive the following while executing the role using the `devel` branch:

    TASK [geerlingguy.docker : Ensure dependencies are installed.] ***********************************************************************************************************************
    [DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying `name: {{ item 
    }}`, please use `name: [u'apt-transport-https', u'ca-certificates']` and remove the loop. This feature will be removed in version 2.11. Deprecation warnings can be disabled by 
    setting deprecation_warnings=False in ansible.cfg.
    ok: [hostname] => (item=[u'apt-transport-https', u'ca-certificates'])

The fix looks trivial, and will actually make the role more efficient. From the `apt` module [notes](https://docs.ansible.com/ansible/latest/modules/apt_module.html#notes):

> When used with a *loop*: each package will be processed individually, it is much more efficient to pass the list directly to the *name* option.

gebn:


geerlingguy:
LGTM

PR #81, opened 2018-07-23
Update README.md

abedk:
Latest docker-compose


geerlingguy:
This would also need to be updated in the defaults: https://github.com/geerlingguy/ansible-role-docker/blob/master/defaults/main.yml#L15

PR #82, opened 2018-07-24
allow downgrade if needed

ryanjfrizzell:
allows downgrade of the docker engine if we need. (EL specific option)


geerlingguy:
The easiest way to downgrade (IMO) is to delete the server or container and start again; I don't usually support downgrades in my playbooks.

PR #96, opened 2018-11-08
add docker_version variable

### LINKED ISSUE #45: Feature request - select docker version
Currently there's no option to specify a specific docker version, which is desired in the case for something like a Kubernetes (which at the moment requires a 17.03.x version), or keeping all installations consistent across a set of servers without necessarily moving them all to the latest version.

At the moment I fixed this by overriding the `docker_package` variable for those specific hosts like this, but this only works for Ubuntu hosts:

```
# Explicitly set the version of docker to be installed to the one here:
docker_version: '17.03.2'
docker_package: 'docker-{{docker_edition}}={{docker_version}}~{{docker_edition}}-0~{{ansible_distribution|lower}}-{{ansible_distribution_release}}'
```

Not sure how this would be handled best, but you need some version specific handling per distribution, since the "version" part of the package name already uses a different separator between `apt` (which uses `=`) and `yum` (which uses `-`).

It would be nice if this could be generalized somehow, but I don't need `yum` support at the moment, so my 'quick hack' works at the moment.

anmoel:
This pull request enables the selection of the docker version.
This also close #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Available variables are listed below, along with default values (see `defaults/m
     docker_edition: 'ce'
     docker_package: "docker-{{ docker_edition }}"
     docker_package_state: present
+    docker_version: "" # like "18.06.0", docker_package_state must be "present", empty means latest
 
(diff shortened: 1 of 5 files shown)


anmoel:
you can set a docker version like '18.06.0' or '18.06'
in the second case it would install the latest patch level

dmperilla:
@anmoel This may impact you, check out this [issue 8039](https://github.com/docker/docker.github.io/issues/8039#issuecomment-461941119) from the docker project.

PR #100, opened 2018-11-21
Added environment proxy_env, become: yes for some tasks

patsevanton:
Hello!
Added environment proxy_env, become: yes for some tasks
https://travis-ci.com/patsevanton/ansible-role-docker/builds/92197253

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,22 @@ None.
     - geerlingguy.docker
 ```
 
+## Example Playbook with proxy
+
+```yaml
+- hosts: all
+  roles:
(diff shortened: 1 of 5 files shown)


geerlingguy:
This role is intended to be run with `become: yes` either applied at the role level or on the playbook level. Also, any proxy settings should be set globally (e.g. either by some other role or in the playbook pre_tasks) instead of being applied via this role itself.

Thank you for the PR, though; but I would rather leave these settings up to the playbook.

patsevanton:
What do you mean proxy settings should be set globally ?

PR #114, opened 2019-01-09
Fix: 17.03.2, 18.09.x and daemon.json

till:
This PR contains:

 - A fix to install older Docker versions, e.g. 17.03.2
 - A fix to allow configuration via `/etc/docker/daemon.json` (systemd script seems oddly restrictive on 18.09.x)


tasks/main.yml
```diff
@@ -5,7 +5,17 @@
 - include_tasks: setup-Debian.yml
   when: ansible_os_family == 'Debian'
 
+- name: Install Docker on CentOS/Redhat
+  when: ansible_os_family == 'RedHat'
+  yum:
+    name:
+      - "docker-ce-selinux-{{ docker_package_version }}"
+      - "{{ docker_package }}-{{ docker_package_version }}"
```
till: This part probably needs some work for when no version is set (and "latest" can be assumed). But this is more or less my first pass at anything complex with ansible, therefor, happy to add to it if anyone finds the time to give me feedback. :)

PR #119, opened 2019-02-07
Adding Fedora31 OS to be tested

froblesmartin:


Sispheor:
this seems indeed needed.

froblesmartin:
@geerlingguy take a look on this pls :+1:

froblesmartin:
@geerlingguy may you review this easy [issue ](https://github.com/geerlingguy/ansible-role-docker/issues/120)+ PR?

geerlingguy:
It sounds like this is only necessary on Fedora—I have tests on ubuntu 18/16, centos 8/7, and debian 10/9 and there are no failures there.

Before merging, I'd like to make sure there are tests for fedora (probably 31, don't need multiple past versions since it doesn't have LTSes like the other distros I test)... The thing is, I'm not sure why the Docker installation doesn't create the `docker` group like it does for every other distro? That seems like it would be a bug with the dnf package rather than something this role would need to add a task for...

nicolas-marcq:
@geerlingguy 
Just tested on a fresh Ubuntu server:
```
docker ps -a
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/json?all=1: dial unix /var/run/docker.sock: connect: permission denied
```

And the docker group is well present.

nicolas-marcq:
Forgot to set the `docker_users`. So on Ubuntu Server 18.04 it's ok.

froblesmartin:
@geerlingguy I have added Fedora31 tests and it works fine. Tried with Fedora29 and got the issue, so I guess it's okay now, even though that change would make the role more robust.

froblesmartin:
Tests green: https://travis-ci.org/github/geerlingguy/ansible-role-docker/builds/668305254

geerlingguy:
Thanks!

PR #122, opened 2019-02-14
systemd isn't very fond of this line

virtualpeter:


virtualpeter:
never mind - ansible keeps putting it back in anyway...

Issue #98, opened 2018-11-13
Remove shim for Docker 18.09.0 support when 18.09.1 or later is released


geerlingguy:
See #97 — basically, to fix that issue I'm adding a shim that fixes running Docker 18.09.0 on any systemd/containerd system (basically anything this role supports besides CentOS 6 and Ubuntu 14.04).

This issue is postponed until the next release of Docker, at which time I'll revert that commit.

geerlingguy:
Heh, "sade". #latenightcommits

Issue #97, opened 2018-11-12
Errors on Ubuntu 16.04 hosts


nre-ableton:
When using this role on a freshly-deployed Ubuntu 16.04 host, we are getting the following error:

```
Unable to restart service docker: Job for docker.service failed because the control process exited with error code. See \"systemctl status docker.service\" and \"journalctl -xe\" for details.\n
```

geerlingguy:
Yeah, I'm also getting that with Docker 18.09 on 18.04, CentOS 7, Debian 9, and basically everywhere. I think something in 18.09 is blowing everything up.

Looking in the logs, I see:

```
Nov 12 22:55:11 instance systemd[1]: Starting containerd container runtime...
-- Subject: Unit containerd.service has begun start-up
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- Unit containerd.service has begun starting up.
Nov 12 22:55:11 instance systemd[1]: Starting Docker Application Container Engine...
-- Subject: Unit docker.service has begun start-up
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- Unit docker.service has begun starting up.
Nov 12 22:55:11 instance systemd[1]: containerd.service: control process exited, code=exited status=1
Nov 12 22:55:11 instance systemd[1]: Failed to start containerd container runtime.
-- Subject: Unit containerd.service has failed
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- Unit containerd.service has failed.
-- 
-- The result is failed.
Nov 12 22:55:11 instance systemd[1]: Dependency failed for Docker Application Container Engine.
-- Subject: Unit docker.service has failed
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- Unit docker.service has failed.
-- 
-- The result is dependency.
Nov 12 22:55:11 instance systemd[1]: Job docker.service/start failed with result 'dependency'.
Nov 12 22:55:11 instance systemd[1]: Unit containerd.service entered failed state.
Nov 12 22:55:11 instance systemd[1]: containerd.service failed.
Nov 12 22:55:11 instance systemd[1]: Stopped Docker Application Container Engine.
-- Subject: Unit docker.service has finished shutting down
-- Defined-By: systemd
[...]

PR #123, opened 2019-02-20
Remove shim for Docker 18.09.0 support when 18.09.1 or later is released

michalklempa:
See: https://docs.docker.com/engine/release-notes/

There are two new releases since #97  fix was introduced, let me propose revert as described in #98


zx1986:
not merge yet :-(

2019/03/25

lucasff:
@geerlingguy ?

KostyaEsmukov:
Given that default `docker_package_state` is `present`, there's still a possibility that this role would be run on a machine with docker 18.09.0, which would still need this shim.

So I believe it would be more robust to keep this shim but don't apply it for the newer versions of docker. Also, perhaps the shim should be explicitly removed for the newer docker versions?

PR #127, opened 2019-03-06
Add Docker Machine

guoqiao:
Currently the role support docker-compose.
Add docker-machine support in the same way.

Signed-off-by: Guo Qiao <guoqiao@gmail.com>


Porkepix:
@geerlingguy What about this PR? Just discovered it, it would have been quite useful to us. It was closed as stale without any other people that commented here.

PR #128, opened 2019-03-11
Skip empty item on docker_users

wainersm:
I wrote a playbook that had something like this:
```
docker_users:
  - "{{ lookup('env','USER') }}"
```
which might return an empty string if running it from inside a container. I noticed ansible-role-docker failed on attempt to add an user with empty name in the docker group. Later I refactory my code to deal with this scenario but I think that It could be avoided by adding a guard on ansible-role-docker directly, once the above code pattern might be often used out there.

### DIFF
--- a/tasks/docker-users.yml
+++ b/tasks/docker-users.yml
@@ -5,3 +5,4 @@
     groups: docker
     append: true
   with_items: "{{ docker_users }}"
+  when: item != ''


tenhishadow:
this is not correct.
ansible-lint will say that it's not good to diff with null string.

it's better to use this:
``` yaml
- name: remove standard kernels if presented
  apt:
    name: "{{ item }}"
    state: absent
    force: yes # bear and mind that it will remove running kernel
    purge: yes # and configs
  loop:
  - "{{ standart_kernels_installed.stdout_lines|default([]) }}"
  when:
  - standart_kernels_installed.stdout_lines | length > 0
  notify:
  - reboot
```

but in general we could just add ansible_user into defaults/main.yml

geerlingguy:
Hmm... I'd have to do some testing before merging something like this; it seems to add some complexity that I don't have the time to consider edge cases for.

PR #136, opened 2019-04-10
u'0' and bare evaluations fixes

freaker2k7:
Fix small bug with quotes causing 0 to be considered as '0', also fixed bare evaluation issue.


geerlingguy:
I believe the bare warnings have been fixed already.

PR #137, opened 2019-04-25
Add proxy into systemd environment

Teebor-Choka:
Proxies were missing in docker daemon causing unreachable network errors.


lechuk47:
+1 To this PR. I need the same functionality.

bellackn:
+1

Teebor-Choka:
> This pull request has been marked 'stale' due to lack of recent activity. If there is no further activity, the PR will be closed in another 30 days. Thank you for your contribution!
> 
> Please read [this blog post](https://www.jeffgeerling.com/blog/2020/enabling-stale-issue-bot-on-my-github-repositories) to see the reasons why I mark pull requests as stale.

Yeah, that's great, dear Mr. bot... but I'd prefer to have this PR actually accepted and merged. Can you tell me, how to achieve that?

geerlingguy:
Configuring proxies (whether for docker or other services) seems like something that would be better done outside this role. Most users of this role don't need to worry about proxies, and adding this functionality adds complexity to the role that would probably apply to more like 20% of the roles users (at best).

Additionally, it's hard to test and maintain this functionality in CI, so it is prone to breakage (especially since none of the infrastructure I use has proxies, so only a small portion of end users would find it broken one day for some inadvertent change).

As with other more advanced docker configuration, I encourage people to put that into either their playbook that runs this role, or into a separate role that configures proxies for everything else.

PR #138, opened 2019-04-29
adds ansible version requirement

wookietreiber:
This role uses both `include_tasks` and `import_tasks`, both of which require Ansible 2.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ An Ansible Role that installs [Docker](https://www.docker.com) on Linux.
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
This is already explicitly defined here: https://github.com/geerlingguy/ansible-role-docker/blob/master/meta/main.yml#L9

The README requirements are more for the purposes of 'already installed on the server' type things, I'm assuming the user will use a modern version of Ansible, and also will be using Galaxy to install the role so the version check is used.

PR #143, opened 2019-05-17
Fixes #141, Add support for Amazon Linux

### LINKED ISSUE #141: Add support to Amazon Linux
As per [documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/docker-basics.html#install_docker) it "only" requires executing
```bash
sudo amazon-linux-extras install docker
```

bkmeneguello:


bkmeneguello:
@geerlingguy you think this PR is OK?

tasks/setup-Amazon.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: Enable Amazon Linux Extras Docker repository
+  command: amazon-linux-extras enable docker  # noqa 301
```
Lusitaniae: Why enable instead of install?

I tried enable, doesn't work.... 

What is the meaning of the comment?

It's also missing `-y` to perform the install automatically instead of waiting for user input.

Lusitaniae:
There's a docker image for Amazon Linux, so we can include it in the test suite.

https://hub.docker.com/_/amazonlinux/

Lusitaniae:
thanks for the PR @bkmeneguello 

I made some comments

tasks/setup-Amazon.yml
bkmeneguello: I've just enabled because it will be installed in later [here](https://github.com/geerlingguy/ansible-role-docker/blob/e697edb53330d3ba3f9cc94978fc0326639cfc2d/tasks/main.yml#L18)

Lusitaniae: does that work? (does it resolve all the dependencies?)

I run install manually now, so not sure.

bkmeneguello: Yes, it does

Lusitaniae: Good to hear, thank you for the feedback

Dmitry1987:
Hi guys, the role is still broken on AWS Linux 2, which version of the role this will be added to?

zenntrix:
This still seems to be a problem, can it be re-opened and addressed?

Issue #124, opened 2019-02-27
Ubuntu seems to require "gnupg" too.


pmav99:
When trying to run the role on a stripped down Ubuntu VM (to be more precise the VM that gets created by [virt-builder](http://libguestfs.org/virt-builder.1.html)) I got the following failure: 
```
$ ansible-playbook kube.yml

PLAY [all] ********************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************
ok: [knode3]
ok: [knode1]
ok: [kmaster1]
ok: [knode2]

TASK [geerlingguy.docker : include_tasks] *************************************************************************************************************************************************************************
skipping: [kmaster1]
skipping: [knode1]
skipping: [knode2]
skipping: [knode3]

TASK [geerlingguy.docker : include_tasks] *************************************************************************************************************************************************************************
included: /home/feanor/.ansible/roles/geerlingguy.docker/tasks/setup-Debian.yml for kmaster1, knode1, knode2, knode3

TASK [geerlingguy.docker : Ensure old versions of Docker are not installed.] **************************************************************************************************************************************
ok: [knode1]
ok: [knode2]
ok: [kmaster1]
ok: [knode3]

TASK [geerlingguy.docker : Ensure dependencies are installed.] ****************************************************************************************************************************************************
changed: [kmaster1]
changed: [knode3]
changed: [knode1]
changed: [knode2]

TASK [geerlingguy.docker : Add Docker apt key.] *******************************************************************************************************************************************************************
fatal: [knode2]: FAILED! => {"changed": false, "msg": "Failed to find required executable gpg in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games"}
...ignoring
fatal: [knode3]: FAILED! => {"changed": false, "msg": "Failed to find required executable gpg in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games"}
...ignoring
fatal: [kmaster1]: FAILED! => {"changed": false, "msg": "Failed to find required executable gpg in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games"}
...ignoring
[...]

PR #148, opened 2019-07-20
install gnupg when not installed ( #124 )

mabeett:
Solves #124 for debian/ubuntu targets


mabeett:
Feel free to comment or indicate corrections for style or logic.

PR #158, opened 2019-09-19
make sure yum isn't turning off docker's network

RLuckom:
Docker relies on the `systemctl` setting `net.ipv4.ip_forward = 1` to give containers access to networks outside the instance. In centos, this setting seems to be on by  default, so every instance has it when it starts up and docker can talk to the world and everything is awesome. But for some reason the `/etc/rc.d/init.d/network` does `sysctl -w net.ipv4.ip_forward=0 > /dev/null 2>&1` when `stop` is called, which it appears to be during some yum updates. This setting never gets changed back to 1 until a reboot occurs, and docker containers lose the ability to talk outside the instance.

### DIFF
--- a/tasks/setup-RedHat.yml
+++ b/tasks/setup-RedHat.yml
@@ -33,3 +33,10 @@
     section: 'docker-{{ docker_edition }}-test'
     option: enabled
     value: '{{ docker_yum_repo_enable_test }}'
+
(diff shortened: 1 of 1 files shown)


RLuckom:
Sorry, meant to commit to a fork

PR #173, opened 2019-11-13
Fix value type warning during repo configure

jonsbun:
Simple fix which allows to avoid the following warning message: "[WARNING]: The value 0 (type int) in a string field was converted to u'0' (type string). If this does not look like what you expect, quote the entire value to ensure it does not change."


nbendafi-yseop:
@jonsbun Following [Ansible Styleguide - Quotes](https://github.com/whitecloud/ansible-styleguide#quotes),
>  The only time you should use double quotes is when they are nested within single quotes (e.g. Jinja map reference), or when your string requires escaping characters (e.g. using "\n" to represent a newline).

So better use `'` instead of `"`

nbendafi-yseop:
:+1:

geerlingguy:
Sorry, didn't notice this PR before I updated this as part of adding support for CentOS 8 earlier today. Thank you for raising the PR though!

PR #174, opened 2019-11-18
Support proxy configuration

weshouman:


weshouman:
Any updates on that?

geerlingguy:
I would recommend you configure any proxy for your playbooks on the playbook level, and it's not something that roles would typically need to worry about, unless there are specific features like proxy configuration within the Docker daemon that have to be accounted for.

weshouman:
I know what you mean, and that's exactly why I opened the pull request here.

The added ```tasks/docker-service.yml``` is role specific proxy setting that should be done so that the task ```Ensure Docker is started and enabled at boot.``` continues successfully, and of course configuring docker should be done after it's installed.

So using the role within a proxified environment requires such edit.

weshouman:
I'm also with the fact that proxy setting should be done on the playbook level, and the added tasks/docker-service.yml allows the role to take the proxy setting into consideration.

IIRC,the role failed to run to completion if the proxy was not set, and of course configuring docker should be done after it's installed.

So using the role within a proxified environment required such edit.

geerlingguy:
Ah, I hadn't looked closely at the PR to notice it was setting `/etc/systemd/system/docker.service.d/proxy.conf`, sorry about that. I'll leave it open and reconsider.

Sispheor:
This Pr is actually usefull I think.
Only a "when" need to be added on the task:
```
- name: Add Docker configuration with proxy set.
  template:
    src: proxy.conf.j2
    dest: /etc/systemd/system/docker.service.d/proxy.conf
  register: proxy_template
```
To be sure we don't apply a proxy on a system that doesn't use it.

tomereli:
> This Pr is actually usefull I think.
> Only a "when" need to be added on the task:
> 
> ```
> - name: Add Docker configuration with proxy set.
>   template:
>     src: proxy.conf.j2
>     dest: /etc/systemd/system/docker.service.d/proxy.conf
>   register: proxy_template
> ```
> 
> To be sure we don't apply a proxy on a system that doesn't use it.

AFAIK since the defaults are empty, systems which doesn't need the proxy will have the config file with empty setting so should work as is.
@geerlingguy what's the status of this PR? It is closed, but is quite useful for everyone behind a corporate proxy..

PR #181, opened 2019-12-26
Add support for sudo, adds become:true for priv tasks

sumkincpp:
Add support for sudo, adds become:true for tasks that require privilege escalation


sumkincpp:
Useless MR, become:True is simpler to use.

PR #140, opened 2019-05-07
Add daemon configuration

henyxia:
Signed-off-by: Jean Wasilewski <henyxia@revs0.com>

PR #108, opened 2018-12-20
Added daemon.json configuration and custom CA

dev-rowbot:
These changes will allow the user to configure the `daemon.json` file during the install as well as add custom Certificate Authorities.

Issue #90, opened 2018-09-27
Feature Request: allow override of dockerd options 


robbyoconnor:
I love your Ansible Roles -- but the major no-go for me is the fact I can't override the storage driver easily.

llbbl:
```
# defaults
docker_default_config:
  storage-driver: overlay2
  log-level: info

# playbook
- name: apply default daemon config
  set_fact:
    docker_json: "{{ docker_default_config | combine(docker_config | default({})) }}"

- name: ensure daemon config file is present
  copy:
    content: |
      {{ docker_json | to_nice_json(indent=2) }}
    dest: /etc/docker/daemon.json
  notify: restart docker

```

how mongrelion is doing it.

llbbl:
FYI devicemapper is depreciated and causing issues. 

https://github.com/docker/cli/pull/1424

stegr04:
This role is installing v18.09.5 for me as of this post. I'm not sure what might have changed since this issue was originally opened but I figured I would share my findings. 

I was looking for the same thing (change storage driver) because docker info was showing it was using devicemapper after installing docker with this role. I found this issue and the solutions discussed/created. Awesome!

I started testing Rickkwa's solution (thank you!). It was working as advertised. Then I started testing without the storage options settings (after manually removing docker). I noticed it was still using overlay2. I wasn't expecting that. I then realized the uninstall wasn't removing the /etc/docker/daemon.json file. 

Good time for some backgroun... all of the hosts I am testing with had docker previously installed. I think the system was setup this way because prior versions of docker required Kernel v4 on CentOs. 

To make sure I wasn't picking up on anything from prior setups I removed /etc/docker and /var/lib/docker. I tested with Rickkwa's again (no storage options again) and it still showed overlay2, remove everything again, tested geerlingguy's role and found it using overlay2. 

To see if there was something else I might have been missing I tested geerlingguy's role on a server I haven't touched except to uninstall all docker packages on the machine. It did not have any /etc/docker/daemon.json file but it did still have /var/lib/docker. When I tested against this node docker info showed me it was using devicemapper. So I did the manual uninstall again, cleaned up the /var/lib/docker and /etc/docker/key.json, ran geerlingguy's role and it's using overlay2.

robbyoconnor:
@geerlingguy

robbyoconnor:
This issue is not resolved

garymoon:
[...]

PR #190, opened 2020-03-18
Add the possibility to customize the options in '/etc/docker/daemon.json'

vulturm:
Hello,
This PR implement the possibility to provide `daemon.json` options in yaml format.
Handy for various customizations.

Please review,
Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,21 +19,21 @@ Available variables are listed below, along with default values (see `defaults/m
 
 The `docker_edition` should be either `ce` (Community Edition) or `ee` (Enterprise Edition). You can also specify a specific version of Docker to install using the distribution-specific format: Red Hat/CentOS: `docker-{{ docker_edition }}-<VERSION>`; Debian/Ubuntu: `docker-{{ docker_edition }}=<VERSION>`.
 
-You can customize `daemon configuration` by providing `daemon.json` values in yaml format. Example:
+You can control whether the package is installed, uninstalled, or at the latest version by setting `docker_package_state` to `present`, `absent`, or `latest`, respectively. Note that the Docker daemon will be automatically restarted if the Docker package is updated. This is a side effect of flushing all handlers (running any of the handlers that have been notified by this and any other role up to this point in the play).
 
-    docker_config_options:
+    docker_service_state: started
+    docker_service_enabled: true
+    docker_restart_handler_state: restarted
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -19,6 +19,16 @@ Available variables are listed below, along with default values (see `defaults/m
 
 The `docker_edition` should be either `ce` (Community Edition) or `ee` (Enterprise Edition). You can also specify a specific version of Docker to install using the distribution-specific format: Red Hat/CentOS: `docker-{{ docker_edition }}-<VERSION>`; Debian/Ubuntu: `docker-{{ docker_edition }}=<VERSION>`.
 
+You can customize `daemon configuration` by providing `daemon.json` values in yaml format. Example:
```
geerlingguy: For my README files, it has the variable first, then the description. So this whole block should be after the "You can control..." paragraph, and the "You can customize..." line needs to be after the variable example. Also, the default should be used here (what is in `defaults/main.yml`).

tasks/main.yml
```diff
@@ -11,6 +11,17 @@
     state: "{{ docker_package_state }}"
   notify: restart docker
 
+- name: Ensure /etc/docker exists
+  file:
+    path: /etc/docker
+    state: directory
+
+- name: Install daemon.json
```
geerlingguy: This task should have a variable to control whether it actually is done or not; some people may already be controlling their `daemon.json` file differently. And since this role doesn't _currently_ manage the file, the default should be to not manage the file.

So, something like `docker_daemon_json_manage: false` in `defaults/main.yml` and documented in the README.

templates/docker-daemon.json.j2
```diff
@@ -0,0 +1 @@
+{{ docker_config_options | default([])| to_nice_json }}
```
geerlingguy: Just thinking through this, let's call the variable `docker_daemon_json_options` instead of `docker_config_options`—it would be a little more directly related to the file under management.

README.md
geerlingguy: This line should also link to the daemon configuration documentation: https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file#daemon-configuration-file

tasks/main.yml
vulturm: What do you think if we don't introduce a new variable and instead, invoke the tasks if `docker_config_options` is defined and length>0 instead ?

vulturm:
Hi @geerlingguy,
any objections against merging this?

Currently this is a blocker for some internal project.
If the code needs to be changed in some how in order to speed up the merge, please suggest.

But, if you're against this idea overall, then I would appreciate if you leave a note here in order to close this and take a different approach internally.

linuxbandit:
This addresses issue #90, and there are other PRs addressing it: #190, #140, #108 
(just wanted to make some order, hopefully one of the three will be picked. Trying to help with issue maintenance)

PR #193, opened 2020-04-07
feat: install docker-compose via control node download

cordula-grau:
This change will download docker-compose from github to the control node and afterwards upload it to the managed server. This is important because servers should not be able to access github.
 
Changes:
* remove delete operation since it is not required with this implementation anymore
* download docker-compose to control node and upload it to the managed server if no version or a version mismatch is detected

### DIFF
--- a/tasks/docker-compose.yml
+++ b/tasks/docker-compose.yml
@@ -5,16 +5,28 @@
   changed_when: false
   failed_when: false
 
-- name: Delete existing docker-compose version if it's different.
-  file:
-    path: "{{ docker_compose_path }}"
-    state: absent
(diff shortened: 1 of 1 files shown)


geerlingguy:
> This is important because servers should not be able to access github.

I use and rely on GitHub for many of my servers' operations, including for many applications cloning repositories and/or pulling updates. I don't think that's a reasonable default for most infrastructure needs.

There are some places where access to GitHub may be intentionally limited, but my role does not cater to that particular need. I'd rather you set `docker_install_compose: false` if you need to install it via an alternative mechanism besides downloading from GitHub—then you can install it however you like in your playbook or in a separate role.

PR #194, opened 2020-04-20
Only install apt-transport-https when necessary

t2d:


geerlingguy:
There's no harm in having the package present on the latest versions of each though, is there? (Especially if it's already present?)

PR #197, opened 2020-04-30
make optional to add the official docker repos

pescobar:
This option allows to use this role in an environment without direct internet access where official docker repos are not accessible and the yum or apt repositories are provided outside of the role. e.g. using katello or a local mirror

The tasks adding the repos are moved to ansible block with a conditional. [it is easier to see here](https://github.com/scicore-unibas-ch/ansible-role-docker/blob/5a35c061731e6a7025ac55a6fc5e320986eae54b/tasks/setup-RedHat.yml#L10-L40)
 
this PR should not change the current default behaviour of the role which is to use the official docker repos


andrew-acpatt:
I was about to do the same thing that you've done, and then I found this PR. Please merge!

pescobar:
any chance to get this PR merged upstream?

geerlingguy:
All my roles assume the server has an Internet connection or at least enough of one to talk to the upstreams—I am on the fence about adding (albeit only a little) complexity to support a use case that I think is very much in the minority of the role's users.

pescobar:
I see your point what I would argue that even when the servers have direct internet access it's common in some corporate environments to use local apt/yum repositories (or mirrors) to reduce bandwidth usage and have more control over what software is presented to the servers.

hyclak:
I'll second @pescobar - my dev hosts do have internet access, but I'm using Katello to mirror repos internally. In test/prod the hosts do not have internet access and are getting packages from Satellite. Would be nice to be able to manage the repo setup separately from the packages in general for these types of use cases.

chrisbalmer:
I also lock down hosts from internet access and mirror what I need to a few select systems. Adds security and reduces data usage for places with data caps (which are definitely still a thing in some places).

PR #200, opened 2020-05-06
support ansible by unprivileged user

genofire:
by using `become: yes`

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,3 +1,4 @@
 ---
 - name: restart docker
+  become: yes
   service: "name=docker state={{ docker_restart_handler_state }}"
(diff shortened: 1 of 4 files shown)


Sleuth56:
Putting this here for anyone like myself that needed to do this as an ansible task instead of a role declaration.
The below ansible code will run the geerlingguy.docker role as root. Assuming that you have not changed the become user ansible variable. `become_user: ` will allow you to change the user. Put that under the `become: yes` line. 
Reference: [Ansible include_role docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_role_module.html) 
```
- name: Install docker
  ansible.builtin.include_role:
    name: geerlingguy.docker
    apply:
      become: yes
```

PR #212, opened 2020-06-18
Add ability to look up the latest docker-compose version from Github API

robinreinhardt:
This small changes add the ability to get the latest docker-compose version from Github API when docker_compose_version is "latest"

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Variables to control the state of the `docker` service, and whether it should st
     docker_compose_version: "1.26.0"
     docker_compose_path: /usr/local/bin/docker-compose
 
-Docker Compose installation options.
+Docker Compose installation options. You can use `docker_compose_version: "latest"` to get the current latest version.
 
     docker_apt_release_channel: stable
     docker_apt_arch: amd64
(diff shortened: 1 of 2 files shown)


geerlingguy:
I like this.

jorianvo:
Please be aware that this code is now broken due to the release of compose v2 i.e. when you now try to get the latest release you'll receive a `404`. This is due to the fact that in compose v2 the assets are renamed  from e.g.
`docker-compose-Linux-x86_64` to `docker-compose-linux-amd64`.

geerlingguy:
Hmm... is there an easy way to work around that?

robinreinhardt:
I've fixed the docker compose download url error

robinreinhardt:
Unfortunately I guess this can not be solved easily.
The download link structure is really messed up across the versions.
Here what I got so far:

version 1: version number without "v", Linux, x86_64
version 2.0.0: version with "v", linux, amd64
since version 2.0.1: version with "v", linux, x86_64

That's why it's not possible to downgrade the version from 2 to 1 like mentioned on #378 

Gladly report if someone can create a logic for this link structure.

Perhaps we should consider deprecating compose version 1 in this role.

PR #216, opened 2020-07-15
debian: update cache pkgs

kaminek:
Signed-off-by: Amine Kherbouche <akherbouche@scaleway.com>

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -13,6 +13,7 @@
       - ca-certificates
       - gnupg2
     state: present
+    update_cache: yes
 
 - name: Add Docker apt key.
   apt_key:


tschmidtke-r7:
I'm seeing failures here as well. Would like to see this merged.

geerlingguy:
Rather than doing this, I recommend people running on Debian/Ubuntu systems add a pre_task that manages the apt caches. Adding update_cache to all package related tasks across my roles would not be as easy and can also lead to slower performance as the cache is updated every time the task is hit.

For an example of pre_task use, see the test playbook that comes with this role: https://github.com/geerlingguy/ansible-role-docker/blob/master/molecule/default/converge.yml#L6-L9

kaminek:
I agree, doing cache updates in every single task is not convenient. I did that directly in the task because It was not redundant, but again a pre_task is a better approach. I'll submit another PR when I find some time.

PR #231, opened 2020-10-05
Add a "docker system prune" cron job

globz-eu:
Hi,
this pull request adds a clean up cron job that performs `docker system prune -af` on a regualr basis.
Please let me know if you have any questions or comments.
Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ before_script:
 script:
   # Run tests.
   - molecule test
+  - molecule test -s clean_up_job
 
 notifications:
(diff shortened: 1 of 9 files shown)


paulkitt:
Would be a great enhancement for the role. Currently adding the same cron job via separate role.

geerlingguy:
This is more of a cleanup task that I would rather not include in the role itself. There are a lot of use cases where you wouldn't want to run a prune regularly and I don't think it follows my own principle of 'meeting the 80% use case'.

I'd recommend managing prune jobs in your own playbook and not in this role directly.

PR #238, opened 2020-11-09
Add bash completion

paulkitt:
Add bash completion for the docker and docker-compose command.

### DIFF
--- a/tasks/docker-compose.yml
+++ b/tasks/docker-compose.yml
@@ -18,3 +18,8 @@
     url: https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-Linux-x86_64
     dest: "{{ docker_compose_path }}"
     mode: 0755
+
+- name: Install bash completion for docker-compose
+  get_url:
+    url: https://raw.githubusercontent.com/docker/compose/{{ docker_compose_version }}/contrib/completion/bash/docker-compose
+    dest: /etc/bash_completion.d/docker-compose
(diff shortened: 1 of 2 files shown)


paulkitt:
Thought long if adding this completion is better placed at a role where shell environments are setup. For the installation of the compose completion the version id is needed. Thats why I thought installing them in the role where the var for the version id is already setup is the best idea.

fourstepper:
I think this is not ideal to be bundled with this role. The problem is that supporting bash is 1) not a primary role of this role (heh) and 2) the target system could not be running bash at all - in which case you are bringing and unnecessary dependency for a lot of users and maintenance costs to the maintainer of the role

paulkitt:
Hallo fourstepper thanks for your comment.
1) the goal of the role is to set up docker. For a lot of people completion is definitely a important part of a running docker env. It is also provided by docker. 
The Completion of docker-compose needs the version number provided as default value in this role. So having everything in on place would be amazing.
2) Yes some maybe but the majority does. I think all systems listed in the [meta file](https://github.com/geerlingguy/ansible-role-docker/blob/master/meta/main.yml) do.

I could make install optional via a variable. Then also other shell envs could be supported if needed. 
What do you think about that?

fourstepper:
Why not make a separate role to fit this use case?

paulkitt:
Like I said before because i think adding completions is part of an installation process. Having them optional is fine for me.

PR #246, opened 2020-12-21
Update cache before install gnupg2 package

orginux:
Hi!
Fixed error when in Debian-like systems used old cache

```
TASK [ansible-role-docker :  Ensure dependencies are installed.]
fatal: [instance]: FAILED! => {"changed": false, "msg": "No package matching 'gnupg2' is available"}

```
Added update cache before install packages

### DIFF
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -4,10 +4,6 @@
   become: true
 
   pre_tasks:
-    - name: Update apt cache.
-      apt: update_cache=yes cache_valid_time=600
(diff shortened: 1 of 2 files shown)


geerlingguy:
I typically recommend managing apt caches on the playbook level, e.g. in a pre_task or something along those lines. Managing apt caches in roles makes things a lot more difficult and leads to a lot of duplicate code and sometimes wasted extra effort.

orginux:
Thanks, good recommendation.
I close this Pull Request and fix the error in my playbook

PR #249, opened 2020-12-27
Fixes #239: Switch from 'Edge' to 'Nightly' option since Edge is not a thing now.

### LINKED ISSUE #239: RedHat/CentOS: No [docker-ce-edge] section in docker-ce.repo.
Hi,

The `docker-ce.repo` file downloaded from Docker does not contain any  `[docker-ce-edge]` section (anymore?). The setting of `docker_yum_repo_enable_edge` then creates an empty section, with neither name nor url.

It is not much of an inconvenience as it does not prevent Docker from installing and setting up correctly. But it makes yum warn about it (`Repository 'docker-ce-edge' is missing name in configuration, using id.`) and makes the option `docker_yum_repo_enable_edge` meaningless.

geerlingguy:
Closes #239.﻿

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,17 +39,17 @@ Docker Compose installation options.
     docker_apt_ignore_key_error: True
     docker_apt_gpg_key: https://download.docker.com/linux/{{ ansible_distribution | lower }}/gpg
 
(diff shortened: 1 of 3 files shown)


geerlingguy:
Rebased with #248 included. Should fix the CentOS 8 build failure.

PR #250, opened 2021-01-04
Add Fedora 32 and Fedora 33 to CI

fourstepper:


geerlingguy:
@fourstepper - For Fedora in particular, I only test on the latest version (just to save CI minutes). Can you modify this PR to only test on `fedora33`?

geerlingguy:
Thanks!

PR #255, opened 2021-02-02
added rootless mode

sluetze:
Hi,

i wrote a small tasklist to introduce rootless mode.
There are some caveats. I only tested ubuntu >20 and also depends on the rpms /debs and is not installed 'directly' via download.
I also don't know if docker-ee packets are named identically

I was not able to enable the daemon in user mode with systemd and created the links and files manually. I'm happy to get some input on how to do it better.

Best Regards
-- sluetze

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -14,6 +14,14 @@ docker_install_compose: true
 docker_compose_version: "1.26.0"
 docker_compose_path: /usr/local/bin/docker-compose
 
+# Rootless Docker options. Systemd only.
+docker_install_rootless: false
+# the service-enabling currently uses symlinks, not systemd
+docker_rootless_service_enabled: true
+# A list of users, which get the docker systemd unitfiles
+docker_rootless_users:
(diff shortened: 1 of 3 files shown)


geerlingguy:
This seems like a nice addition.

espro:
This is not ready to be merged, contains personal info, is not compatible with RHEL, etc.

sluetze:
@espro thanks, found that personal info part, can you tell me what is incompatible with RHEL?

espro:
* `newuidmap` and `newgidmap` are not provided by `uidmap` package on RHEL, they are in `shadow-utils` and only available starting with RHEL 7.7
* https://docs.docker.com/engine/security/rootless/ suggests there are other package requirements, and if using RHEL 8 with SELinux enabled changes to the policy may be required (and the recommendation in the Docker docs is lazy)

There could be other issues, so proper testing is required and since this role is aimed at Debian and Redhat, I think ensuring compatibility is a requirement

PR #258, opened 2021-02-09
Ensure Podman and Buildah are not installed on RHEL-based systems

NlVlN:
Presence of buildah and podman creates version issue for runc and makes the role fail on fresh centos 8 installation.


rausub:
can this be reopened and merged, similar issue in oracle

PR #260, opened 2021-02-23
working and tested daemon.json settings

msirovy:


nickbroon:
Is this superseded by #261 ?

geerlingguy:
Closing since I also see #261 has the same commits (and then some).

PR #262, opened 2021-03-04
add possibility to hold docker package

vDMG:
Hello,

Thanks for the amazing work.

I would like to make a proposal to add a feature to help people adding a hold on docker package once installed.

Indeed, we have a use cases where we want to hold docker once installed to avoid unwanted upgrades running `apt-get upgrade` for example. Like many people, we want to maintain docker only using this role, and having the possibility to hold the docker package is an interesting feature.

For now, i have not include RedHat solutions for the hold because it seems really complex to hold a package (needs to install a specific plugin). But if someone have a solution i'm glad to edit this PR.
Moreover, i have not move my modifications on the debian file because there is no logic for post-install action. But i can make the necessary work if you want

I'm ready to hear from your feedbacks and make modifications on my PR.

Thank you.


BarthV:
This feature would be awesome ! :+1: 

Is it possible to rebase & review it please :)

PR #265, opened 2021-03-12
Add ability to configure package mirrors

slicen:
These changes are needed for the role to be usable in an offline environment where github.com and docker.com are inaccessible, or if a user desires to use a local mirror. It allows for configuration of the docker-compose download URL, as well as setting a YUM and APT package mirror. Finally, docker-compose is only downloaded when needed.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -16,17 +16,18 @@ docker_compose_url: https://github.com/docker/compose/releases/download/{{ docke
 docker_compose_path: /usr/local/bin/docker-compose
 
 # Used only for Debian/Ubuntu. Switch 'stable' to 'nightly' if needed.
+docker_repo_url: https://download.docker.com/linux
 docker_apt_release_channel: stable
 docker_apt_arch: amd64
-docker_apt_repository: "deb [arch={{ docker_apt_arch }}] https://download.docker.com/linux/{{ ansible_distribution | lower }} {{ ansible_distribution_release }} {{ docker_apt_release_channel }}"
+docker_apt_repository: "deb [arch={{ docker_apt_arch }}] {{ docker_repo_url }}/{{ ansible_distribution | lower }} {{ ansible_distribution_release }} {{ docker_apt_release_channel }}"
 docker_apt_ignore_key_error: true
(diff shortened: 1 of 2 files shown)


defaults/main.yml
```diff
@@ -12,20 +12,22 @@ docker_restart_handler_state: restarted
 # Docker Compose options.
 docker_install_compose: true
 docker_compose_version: "1.26.0"
+docker_compose_url: https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-Linux-x86_64
 docker_compose_path: /usr/local/bin/docker-compose
 
 # Used only for Debian/Ubuntu. Switch 'stable' to 'nightly' if needed.
+docker_url: https://download.docker.com/linux
```
geerlingguy, line 19: Can you name this `docker_repo_url` or maybe `docker_download_url`? I'm just thinking long term if there is some other feature that requires a download from Docker that uses a different URL, I don't want the more generic `docker_url` taken.

tasks/docker-compose.yml
```diff
@@ -15,6 +15,9 @@
 
 - name: Install Docker Compose (if configured).
   get_url:
-    url: https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-Linux-x86_64
+    url: "{{ docker_compose_url }}"
     dest: "{{ docker_compose_path }}"
     mode: 0755
+  when: >
+    docker_compose_current_version.stdout is not defined
+    or docker_compose_version not in docker_compose_current_version.stdout
```
geerlingguy, line 23: Just to be clear, this is to make it so this task is _only_ run at all if the docker-compose command shows the version differs from the version desired?

defaults/main.yml
slicen: Good idea. Have amended this commit to use `docker_repo_url` instead.

tasks/docker-compose.yml
slicen: That's correct, @geerlingguy! I observed that prior to this commit, docker-compose was re-downloaded on subsequent runs despite already having the desired version (not idempotent). I have amended the commit for clarity, as the `docker-compose --version` task was also not previously constrained to use the docker-compose binary at `docker_compose_path`.

PR #268, opened 2021-03-23
Do not install gnupg2 on Ubuntu 20.04 and superior

toxinu:
This pull request is about adding a condition to not install `gnupg2` on Ubuntu 20.04 and superior since this package is no more.

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -11,8 +11,13 @@
     name:
       - apt-transport-https
       - ca-certificates
-      - gnupg2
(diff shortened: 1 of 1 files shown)


geerlingguy:
This PR actually breaks the ubuntu 20.04 build...

toxinu:
> This PR actually breaks the ubuntu 20.04 build...

Should be fixed now. 👍

BarthV:
:+1:

IanoNjuguna:
💪🏿

PR #271, opened 2021-03-30
More dynamic `docker_apt_arch` for Debian

kennethkalmer:
Based on the `nomad_architecture_map` variable from brianshumate.nomad, allows for a mixture of architecture families.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -18,7 +18,14 @@ docker_compose_path: /usr/local/bin/docker-compose
 # Used only for Debian/Ubuntu. Switch 'stable' to 'nightly' if needed.
 docker_repo_url: https://download.docker.com/linux
 docker_apt_release_channel: stable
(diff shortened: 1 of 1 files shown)


fourstepper:
+1 this looks great

geerlingguy:
@kennethkalmer - Can you give more context about scenarios where this comes into play?

kennethkalmer:
> @kennethkalmer - Can you give more context about scenarios where this comes into play?

Deploying to architectures other than `amd64`, like AWS Graviton instances (`a1.*`) or I imagine Raspberry PI’s

geerlingguy:
@kennethkalmer - Ah... for non-amd64, I actually have a separate role: https://github.com/geerlingguy/ansible-role-docker_arm (mostly because there are enough differences between the arm install and the amd64 that it was annoying trying to keep them together in one role.

PR #275, opened 2021-04-20
Explicitly define the package repo filename

eengstrom:
On debian-flavored systems, failure to do this means that ansible will generate the repository filename (in `/etc/apt/sources.list.d/`), and that filename will include many elements of the repository, which can include distribution name and version numbers.  This means that after upgrading distribution versions, one will end up with multiple repository definitions in separate files rather than updating the same file on each playbook/role execution.

I also updated the naming for RedHat systems, but no behavior change - only the ability to have the user define the filename via variable.


eengstrom:
@geerlingguy - thanks for merging the other PR I submitted.  I also created three other minor-change PRs, including this one, but I did not create issues for each of them.  Would you prefer an issue associated with each?

PR #276, opened 2021-04-20
Enable/Disable the (systemd) `docker.service` and `.socket`

### LINKED ISSUE #279: should enable/disable both the docker.service and docker.socket
Currently the role disables only the `docker.service`, but if someone attempts to connect to the socket, the service will be auto-started.  This becomes a bigger deal when using `docker-rootless`, as the recommendation is to disable the system-level service/socket.

Trivial fix, in PR #276

eengstrom:
When using this role to install `docker` but only intend for it's use in 
`rootless` mode, the Docker documentation suggests disabling 
`docker.service`.

However, that also leaves the `docker.socket` running, which whenever it 
receieves a connect, the daemon (service) is restarted.  

This patch enables/disables the both the service and the socket.

Resolves #279

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -11,12 +11,15 @@
     state: "{{ docker_package_state }}"
   notify: restart docker
 
-- name: Ensure Docker is started and enabled at boot.
+- name: Configure docker service and socket state
   service:
-    name: docker
+    name: "docker.{{ item }}"
     state: "{{ docker_service_state }}"
     enabled: "{{ docker_service_enabled }}"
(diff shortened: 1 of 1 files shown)


eengstrom:
Updated this branch off of current `master`, and added corresponding issue #279.

eengstrom:
@geerlingguy - does this pass the "Jeff"-test ? ;-). 

Seriously, I'm somewhat hopeful you can find some time to look at this and the other two PRs I submitted, and if you need any additional rationale or whatever, please ask.

PR #277, opened 2021-04-20
Retry `docker` package installation at least once.

eengstrom:
Why: when a previous installation was removed, but left behind service 
configurations, and the docker service is disabled, package installation 
will occasionally fail the first time.

This occurs under Ubuntu 18.04, minimally.


fourstepper:
That's interesting. What is the error it gives?

PR #278, opened 2021-05-16
Update ci.yml

fourstepper:
Fedora 34 is out - let's use that for testing from now on


geerlingguy:
Annoyingly, GitHub Actions seems to be having some issues today. I can't get the run to start: https://github.com/geerlingguy/ansible-role-docker/actions/runs/846733250

geerlingguy:
K, got it triggered by manually cancelling/re-running.

PR #284, opened 2021-07-27
Add support for Photon OS

paladin-devops:
Install Docker with tdnf for Photon OS.


geerlingguy:
I'd rather someone maintain a fork with tested support for Photon OS, since I don't use it, can't test it in CI, and it looks like it also uses non-standard package naming conventions.

Issue #286, opened 2021-08-07
update_cache required


profhase:
On Debian install I get:
```
fatal: [molecule-test]: FAILED! => {"changed": false, "msg": "No package matching 'gnupg2' is available"}
```
fix:
use `update_cache: yes`

PR #293, opened 2021-08-17
Debian Install : move `update_cache` on top

profhase:
fixes https://github.com/geerlingguy/ansible-role-docker/issues/286

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -12,6 +12,7 @@
       - apt-transport-https
       - ca-certificates
     state: present
+    update_cache: true
 
 - name: Ensure additional dependencies are installed (on Ubuntu < 20.04 and any other systems).
   apt:


geerlingguy:
@profhase - I don't do this in my roles since apt cache management should happen on the playbook level (since if I had every role manage caches it would mess up idempotence and add a lot of overhead to playbook runs with multiple roles).

profhase:
@geerlingguy Thank you for the Explanation. I see your point - there is a cache update (after adding docker repo, though).

I don't see how it would break the idempotence. Would it be a way to set a `cache_valid_time` at ~10mins (As in `converge.yml`)?

Otherwise the role fails on a new machine. IMHO there should be at least a hint in the Readme

geerlingguy:
@profhase - It's something that you kind of get in the habit of doing with all Ansible playbooks if you use Debian. Just like in the converge playbook, I always lead my playbooks with one task (usually in `pre_tasks`) that updates apt caches, then I do the rest of the work on the server.

I don't add it to individual roles because I maintain over a hundred of them, and to be consistent I'd have to add that task to all 100 roles, which would mean on some builds, dozens of extra tasks that do nothing. Package cache management should be done on the playbook level, not in roles.

Issue #285, opened 2021-08-01
apt-transport-https was already set to manually installed.


corburn:
os: debian sid
role commit: e5c8040aff90e7d8e540bc1439579bc2f78a6543

fatal: [node]: FAILED! => {"changed": false, "msg": "'/usr/bin/apt-mark manual apt-transport-https ca-certificates' failed: E: Could not create temporary file for /var/lib/apt/extended_states - mkstemp (13: Permission denied)\nE: Failed to write temporary StateFile /var/lib/apt/extended_states\n", "rc": 100, "stderr": "E: Could not create temporary file for /var/lib/apt/extended_states - mkstemp (13: Permission denied)\nE: Failed to write temporary StateFile /var/lib/apt/extended_states\n", "stderr_lines": ["E: Could not create temporary file for /var/lib/apt/extended_states - mkstemp (13: Permission denied)", "E: Failed to write temporary StateFile /var/lib/apt/extended_states"], "stdout": "apt-transport-https was already set to manually installed.\n", "stdout_lines": ["apt-transport-https was already set to manually installed."]}

gee-forr:
Howdy...

I just encountered this myself. It's erroring out because the user doesn't have permission to install anything. I run my playbooks as a `deploy` user, and then switch to root ad-hoc as needed.

Call the role in your playbook like so, and you should be good:

```yaml
    - { role: geerlingguy.docker, tags: docker, become: true }
```

The `become: true` runs the entire role as root (assuming your user can sudo). Hope this helps.

staticdev:
@gee-forr I don't think running the whole role as root is recommended from security point-of-view. Although I see in the code, for instance, `apt` steps don't have `become: true`, so probably this is the only way around. I see an inconsistency here, since the documentation put as example playbook using this role without become.

Maybe a PR would be needed to be able this role as a non-root user or adding `become: true` to the example playbook, if this is intentional. Any thoughts on that @geerlingguy?

gee-forr:
> @gee-forr I don't think running the whole role as root is recommended from security point-of-view.

100% agree - I only switch to root on an as-needs basis.

The docs don't use become, because perhaps @geerlingguy builds all his roles to implicitly run as root?

geerlingguy:
[...]

PR #296, opened 2021-08-29
Add README note re: role permissions

gee-forr:
Addresses https://github.com/geerlingguy/ansible-role-docker/issues/285

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ An Ansible Role that installs [Docker](https://www.docker.com) on Linux.
 
 ## Requirements
 
-None.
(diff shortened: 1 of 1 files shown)


gee-forr:
@geerlingguy - any chance of merging this tiny change in before stale bot sends it to the big farm upstate to play with the other old PR's?

geerlingguy:
I don't have that set for any of my other roles—it's implied that my roles are run as a user with sufficient permissions to install and manage software (e.g. `become: true`, and my example code usually highlights that requirement (https://github.com/geerlingguy/ansible-role-docker#example-playbook) — I just noticed the example playbook for this role doesn't have it set, so that should probably be fixed instead.

PR #301, opened 2021-09-11
feat: docker-compose command-line completion

webarchitect609:
Added:

- docker-compose comman-line completion installation.

Updated:

- docker-compose updated from 1.26.0 to 1.29.2.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -14,8 +14,8 @@ docker_install_compose: true
 docker_compose_version: "1.29.2"
 docker_compose_url: https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-Linux-x86_64
 docker_compose_path: /usr/local/bin/docker-compose
-docker_compose_completion_url: https://raw.githubusercontent.com/docker/compose/{{ docker_compose_version }}/contrib/completion/bash/docker-compose
-docker_compose_completion_path: /etc/bash_completion.d/docker-compose
+docker_compose_bash_completion_url: https://raw.githubusercontent.com/docker/compose/{{ docker_compose_version }}/contrib/completion/bash/docker-compose
+docker_compose_bash_completion_path: /etc/bash_completion.d/docker-compose
 
 # Docker repo URL.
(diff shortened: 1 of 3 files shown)


defaults/main.yml
```diff
@@ -11,9 +11,11 @@ docker_restart_handler_state: restarted
 
 # Docker Compose options.
 docker_install_compose: true
-docker_compose_version: "1.26.0"
+docker_compose_version: "1.29.2"
 docker_compose_url: https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-Linux-x86_64
 docker_compose_path: /usr/local/bin/docker-compose
+docker_compose_completion_url: https://raw.githubusercontent.com/docker/compose/{{ docker_compose_version }}/contrib/completion/bash/docker-compose
+docker_compose_completion_path: /etc/bash_completion.d/docker-compose
```
geerlingguy, line 18: This seems like it would also be dependent on the user having bash completion installed, and would also only target bash users, correct? Or would this path be universally-applicable if running ZSH, Fish, etc?

webarchitect609: Yes, this path is only for `bash` with bash completion installed. For ZSH there's another script with other algorythm of installation.

So I think I have to check first if bash completion is installed and then apply this script. Regarding ZSH installation is more complicated and I'm not familiar with ZSH.

For Fish no information in docs. Probably, it's not supported.

Docs: [Command-line completion](https://docs.docker.com/compose/completion/)

webarchitect609: I've updated the PR: check for 'bash-completion' **before** installing docker-compose completion. So bash users will be lucky, others — will remain unharmed :)


Honestly tested:

1) by `molecule test`
2) by `molecule test` + installing `bash-completion` on the fly('cause docker image doesn't have it)
3) on my real laptop with `Ubuntu 20.04 LTS`
4) on my other real laptop with `Ubuntu 18.04 LTS`

webarchitect609: @geerlingguy , what do you think of my PR now? Is there any more critics about it?

PR #311, opened 2021-10-30
Remove sudo dependency

fauust:
The sudo dependency should not be needed because sudo escalation is
managed by Ansible.


geerlingguy:
True, thanks!

Issue #121, opened 2019-02-07
Feature request: allow Docker downgrade


donhector:
### Use case
You deploy Docker to your fleet using this role and its defaults. This will deploy the latest version of Docker available in the platform repos (in my case, I'm running Ubuntu 16.04 and I'm getting 18.09.1 at the time of writing). Then you find a bug in 18.09.1 and want to downgrade to an earlier version, also available in the platform repo, for example: 18.06.1

So I re-ran the role but this time overriding the `docker_package` variable pinning it to an earlier package version like this: `docker_package: docker-ce=18.06.1~ce~3-0~ubuntu`

This is the error I got:

```
TASK [geerlingguy.docker : include_tasks] **********************************************************************************************************************************************************************************************************************************
skipping: [127.0.0.1]

TASK [geerlingguy.docker : include_tasks] **********************************************************************************************************************************************************************************************************************************
included: /exports/nfs/workdir/business-events/.bootstrap/ansible/roles/galaxy/geerlingguy.docker/tasks/setup-Debian.yml for 127.0.0.1

TASK [geerlingguy.docker : Ensure old versions of Docker are not installed.] ***********************************************************************************************************************************************************************************************
ok: [127.0.0.1]

TASK [geerlingguy.docker : Ensure dependencies are installed.] *************************************************************************************************************************************************************************************************************
ok: [127.0.0.1]

TASK [geerlingguy.docker : Add Docker apt key.] ****************************************************************************************************************************************************************************************************************************
ok: [127.0.0.1]

TASK [geerlingguy.docker : Ensure curl is present (on older systems without SNI).] *****************************************************************************************************************************************************************************************
skipping: [127.0.0.1]

TASK [geerlingguy.docker : Add Docker apt key (alternative for older systems without SNI).] ********************************************************************************************************************************************************************************
skipping: [127.0.0.1]

TASK [geerlingguy.docker : Add Docker repository.] *************************************************************************************************************************************************************************************************************************
ok: [127.0.0.1]

TASK [geerlingguy.docker : Install Docker.] ********************************************************************************************************************************************************************************************************************************
[...]

PR #317, opened 2021-11-24
Allow package downgrading

amotl:
Dear Jeff,

thanks a stack for this excellent Ansible Role. As we are currently ramping up our new infrastructure based on Ansible, we ran into #121. This patch attempts to improve the situation.

With kind regards,
Andreas.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -5,12 +5,22 @@
 - include_tasks: setup-Debian.yml
   when: ansible_os_family == 'Debian'
 
-- name: Install Docker.
+- name: Install Docker (Ansible <2.12).
   package:
     name: "{{ docker_package }}"
     state: "{{ docker_package_state }}"
   notify: restart docker
   ignore_errors: "{{ ansible_check_mode }}"
+  when: "ansible_version.full is version_compare('2.12', '<')"
+
+- name: Install Docker (Ansible >=2.12).
+  package:
+    name: "{{ docker_package }}"
+    state: "{{ docker_package_state }}"
+    allow_downgrade: true
+  notify: restart docker
+  ignore_errors: "{{ ansible_check_mode }}"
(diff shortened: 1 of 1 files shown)


tasks/main.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- name: Verify Ansible is modern enough to support `allow_downgrade` for yum+apt.
+  assert:
+    that: "ansible_version.full is version_compare('2.12', '>=')"
```
geerlingguy, line 4: The difficulty here is there are a _lot_ of users still running 2.9, and probably a good number running 2.10 and 2.11 in between, who would immediately be blocked from using this role if we go this route...

amotl: Ah right, I kind of expected this. So, I believe the best thing is to put this patch back into draft mode?

If you see any other chance, to conditionally implement that feature for users already on `ansible-core>=2.12`, instead of croaking on the assertion, please let me know. As I am a newcomer to Ansible, I would need some guidance. On the other hand, I believe it would probably complicate things too much.

amotl: P.S.: I also updated the patch to adjust `galaxy_info.min_ansible_version: 2.12` within `meta/main.yml` instead.

amotl: Hi again,

based on your feedback, and learning some more bits of Ansible, I took another approach and updated this patch to use conditional dispatching between two tasks, responsible for installing the package, based on the version of `ansible-core`. In order to move on, I believe many roles and collections on Galaxy will start using that approach.

Earlier, I said:

> I believe [such a patch] would probably complicate things too much.

When looking at the patch now, I believe the impact is reasonable and solves an important detail for us and others, being able to roll back to previous package versions, even on `apt`-based systems.

With kind regards,
Andreas.

PR #319, opened 2021-12-23
Extracted `docker_compose_arch` variable from `docker_compose_url`

OrBin:


geerlingguy:
Does this print out the correct arch for all platforms that are necessary (e.g. `x86_64` in that format, and I would imagine `aarch64` or `arm64`?).

Note that I maintain a separate role for arm64: https://github.com/geerlingguy/ansible-role-docker_arm

OrBin:
I'm not sure I understand you question - what do you exactly mean by "print out the correct arch"?
I just thought it would be a good idea to have the arch configurable separately from the URL, as you did with `docker_apt_arch`.

geerlingguy:
@OrBin - Ah, in that case, can you also make sure to add a line of documentation for the new default variable in the README? Then I'd be happy to merge.

OrBin:
Sure, I will when I have some free time. Thanks.

OrBin:
@geerlingguy - done, sorry it took so long. Does it look good to you?

geerlingguy:
Thanks!

PR #320, opened 2021-12-31
Add OS specific variable overrides.

adamus1red:
Specific example is Alpine where the [repos](https://pkgs.alpinelinux.org/packages?name=docker&branch=edge) use `docker` rather than `docker-ce` or `docker-ee`

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,4 +1,14 @@
 ---
+- name: Load OS specific package information
+  include_vars: "{{ lookup('first_found', params) }}"
+  vars:
+    params:
+      files:
+        - '{{ansible_distribution}}.yml'
+        - '{{ansible_os_family}}.yml'
+        - default.yml
+      paths:
+        - 'vars'
 - include_tasks: setup-RedHat.yml
   when: ansible_os_family == 'RedHat'
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
-- name: Load OS specific package information
+- name: Load OS-specific vars.
   include_vars: "{{ lookup('first_found', params) }}"
   vars:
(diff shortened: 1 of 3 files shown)


adamantike:
In case this approach gets approved, it could also include an `Archlinux.yml` file with `docker_package: "docker"` too.

adamantike:
I would really appreciate if this can be reviewed, considering that is the only missing change for this role to work out of the box with Arch Linux based distros (otherwise, `docker_package` has to be set to `"docker"`).

tasks/main.yml
```diff
@@ -1,4 +1,14 @@
 ---
+- name: Load OS specific package information
```
geerlingguy, line 2:
```suggestion
- name: Load OS-specific vars.
```

tasks/main.yml
```diff
@@ -1,4 +1,14 @@
...
+  include_vars: "{{ lookup('first_found', params) }}"
+  vars:
+    params:
+      files:
+        - '{{ansible_distribution}}.yml'
+        - '{{ansible_os_family}}.yml'
+        - default.yml
+      paths:
+        - 'vars'
```
geerlingguy, line 11:
```suggestion
        - 'vars'

```

geerlingguy:
I've merged this in to make it more convenient for Arch/Alpine users, but note that precedence means it could be harder for end users to override any vars defined in these included files. So use them sparingly—it makes the role a tiny bit less flexible while also making some platforms better supported OOTB.

@adamantike - if you want to add in a separate PR to add Arch support I'd be willing to merge it (make sure to add to meta/main.yml the new platform, and make sure the name of the file is correct with correct capitalization (not sure if `ArchLinux`, `Arch`, or `Archlinux`)).

PR #323, opened 2022-01-05
added apt update to debian setup

bigsmokecriminal:
I had troubles installing docker today.(Ubuntu 20.04). Interestingly, I had no problems on a second machine which actually was identical (both vagrants). I'm not sure how to reproduce tho...

`FAILED! => {"changed": false, "msg": "No package matching 'docker-ce' is available"}`

Went onto the machine and apt cache did not know docker. A small apt update doesn't hurt and would have fixed that.

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -49,3 +49,7 @@
     repo: "{{ docker_apt_repository }}"
     state: present
     update_cache: true
+    
+- name: Update apt cache
+  apt:
+    update_cache: yes


geerlingguy:
This change would break idempotence since it would update every time—I recommend people add any apt cache management on the playbook level, otherwise I would have to add some sort of cache update to every single role I maintain, and it would add a lot of overhead for playbooks using a bunch of roles.

See how it's done in the converge playbook for an example: https://github.com/geerlingguy/ansible-role-docker/blob/master/molecule/default/converge.yml#L6-L9

PR #324, opened 2022-01-07
Add docker_service_environment variable to allow environment variables such as HTTP_PROXY.

CoaxVex:
Add a docker_service_environment variable, which allows setting additional environment variables to the systemd service, such as HTTP_PROXY, HTTPS_PROXY and NO_PROXY.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -27,6 +27,23 @@
   when: docker_daemon_options.keys() | length > 0
   notify: restart docker
 
+- name: Ensure systemd drop-in directory exists
+  file:
+    path: /etc/systemd/system/docker.service.d
+    state: directory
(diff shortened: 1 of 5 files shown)


tasks/main.yml
```diff
@@ -27,6 +27,23 @@
   when: docker_daemon_options.keys() | length > 0
   notify: restart docker
 
+- name: Ensure systemd drop-in directory exists
+  file:
+    path: /etc/systemd/system/docker.service.d
```
b1tray3r, line 32: This hard coded path could be a problem on systems with custom configurations. 
It should be changeable like "{{ DOCKER_SERVICE_D_PATH|default("/etc/systemd/system/docker.service.d") }}".

Issue #327, opened 2022-01-13
ansible-lint yaml warning: risky-shell-pipe: Shells that use pipes should set the pipefail option


kare:
I have `ansible-role-docker` installed in my playbook and when I run command `ansible-lint playbook.yaml` I get the following warning:

```
risky-shell-pipe: Shells that use pipes should set the pipefail option
tasks/setup-Debian.yml:40 Task/Handler: Add Docker apt key (alternative for older systems without SNI).
```
I don't expect to see any warnings from `ansible-lint`.

PR #328, opened 2022-01-13
refactor: add error check for shell pipe

kare:
```
risky-shell-pipe: Shells that use pipes should set the pipefail option
tasks/setup-Debian.yml:40 Task/Handler: Add Docker apt key (alternative for older systems without SNI).
```
Fixes geerlingguy/ansible-role-docker#327

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -39,6 +39,7 @@
 
 - name: Add Docker apt key (alternative for older systems without SNI).
   shell: >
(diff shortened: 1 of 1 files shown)


geerlingguy:
This does not work cross platform, and I've already accounted for it in https://github.com/geerlingguy/ansible-role-docker/blob/master/.ansible-lint#L3 — please see https://github.com/ansible-community/ansible-lint/issues/497

Issue #304, opened 2021-10-08
Support Docker Compose v2


fabricat:
As stated on https://github.com/docker/compose/blob/v2/README.md:

> The installation instructions for Compose V2 differ from V1.
> V2 is not a standalone binary anymore and installation scripts will have to be adjusted.

As an addition, you could also consider installing [compose-switch](https://github.com/docker/compose-switch) to translate V1's `docker-compose ...` commands into V2's `docker compose ...`.

geerlingguy:
Not stale I don't think.

JMLX42:
I can confirm that setting `docker_compose_version: 'v2.0.1'` works as expected.

geerlingguy:
All right, I updated the default, tweaked the download URL to use lowercase `linux` by default, and I think that closes this issue.

PR #329, opened 2022-01-15
update for arm and docker-compose v2 and above

Lithimlin:
Role now is able to handle arm architectures like from a Raspberry Pi. This includes automatic architecture detection.
Also includes capabilities of installing docker-compose versions 2 and above.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -11,8 +11,8 @@ docker_restart_handler_state: restarted
 
 # Docker Compose options.
 docker_install_compose: true
-docker_compose_version: "1.26.0"
-docker_compose_url: https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-Linux-x86_64
+docker_compose_version: "2.2.3"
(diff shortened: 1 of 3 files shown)


geerlingguy:
This PR seems to be adding two features which makes it a little less likely to get merged (one of the two at least)—I just merged a PR to at least allow some arch-specific configuration to be changed (see #319), and please see https://github.com/geerlingguy/ansible-role-docker/issues/304 for discussion about Compose v2 (which I'd rather add in its own PR too).

Lithimlin:
Ah, I hadn't realized that the commands of compose v2 changed so much since I installed it a while ago and also used the compose-switch. Since compose v2 is not (yet) supported by the ansible docker-compose module though it might be questionable to install it anyway. I'll split this PR into two and probably add a note to the README addressing this issue.

https://github.com/geerlingguy/ansible-role-docker/issues/304#issuecomment-945557008 hints at an issue with my automatic architecture implementation. I don't have a Pi Zero so I wasn't able to test this with an armv6 and it seems there is an extra `1` there. I'll still open up a PR though and ask people to test it.

All this might take a few days.

Lithimlin:
I've split this PR into two (#330 and #331).
The compose v2 support is not yet added as it would profit from either the compose-switch or the compose docker plugin (see https://github.com/geerlingguy/ansible-role-docker/issues/304#issuecomment-944468757)

PR #330, opened 2022-01-26
Arch detect

Lithimlin:
Split of #329:
> Role now is able to handle arm architectures like from a Raspberry Pi. This includes automatic architecture detection.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -12,7 +12,6 @@ docker_restart_handler_state: restarted
 # Docker Compose options.
 docker_install_compose: true
 docker_compose_version: "1.26.0"
-docker_compose_arch: x86_64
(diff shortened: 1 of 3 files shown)


Lithimlin:
Update: now also includes the automatic architecture for docker-compose

MrSuicideParrot:
I think this pull request could be improve removing dpkg dependence.
According to the source.list documentation you can use  something like `$(ARCH)`, which will expand to the platform architecture.
>     suite may also contain a variable, $(ARCH) which expands to the Debian architecture (such as amd64 or armel) used on the system. This permits architecture-independent sources.list files to be used. In general this is only of interest when specifying an exact path; APT will automatically generate a URI with the current architecture otherwise. 
Source: http://manpages.ubuntu.com/manpages/jammy/en/man5/sources.list.5.html

PR #344, opened 2022-04-15
Update setup-Debian.yml

modem7:
Update to remove apt_key.

This module is deprecated, and this is the [Debian guideline](https://wiki.debian.org/DebianRepository/UseThirdParty) workaround until https://github.com/ansible/ansible/pull/77340 is released.

Further info [here](https://www.reddit.com/r/ansible/comments/u2vy26/comment/i4pbgj9/)

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -4,6 +4,9 @@
     name:
       - docker
       - docker-engine
+      - docker.io
+      - containerd
+      - runc
     state: absent
 
 - name: Ensure dependencies are installed.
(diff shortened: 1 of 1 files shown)


geerlingguy:
I would rather not remove something just for being deprecated when the replacement option is a bit more unmaintainable using a shell command. I'd rather wait until we see what Ansible chooses to do, whether it's to modify the `apt_key` module to work in the new system, or to create a new module for it.

modem7:
That's fair!

Hopefully the replacement will be made live at some point soon!

PR #348, opened 2022-05-01
Override Systemd conf when hosts is defined

Ripolin:
When ```hosts``` property is defined in ```daemon.json``` it creates a conflict whith default ExecStart from Systemd configuration (```/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock```).

With this PR, when ```hosts``` is defined in var ```docker_daemon_options```, Systemd configuration is patch to not use ```-H``` arg.

### DIFF
--- a/files/systemd-override.conf
+++ b/files/systemd-override.conf
@@ -1,0 +1,3 @@
+[Service]
+ExecStart=
+ExecStart=/usr/bin/dockerd --containerd=/run/containerd/containerd.sock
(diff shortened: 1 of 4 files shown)


Ripolin:
FYI I have rebased my fork https://github.com/Ripolin/ansible-role-docker/tree/systemd-override from master branch. Perhaps someday @geerlingguy will reopen this PR and merge it, wait and see ...

PR #349, opened 2022-05-03
Use with Ansible section work with minimal hosts

mhkarimi1383:
Some times when I get VM from some providers some extra packages are needed in order to make it working specially when OS is a little bit minimal

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,13 @@ Many users of this role wish to also use Ansible to then _build_ Docker images a
 
   vars:
     pip_install_packages:
+      - name: wheel
+      - name: pip
+        state: latest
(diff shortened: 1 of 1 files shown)


geerlingguy:
@mhkarimi1383 - Since those aren't generally required with the OS installs I see, I'd rather not include them here, but rather have them added in in your own playbook.

mhkarimi1383:
But this is the result of searching, finding and facing with a lot  of errors

This could be added to the example(s) for others too

geerlingguy:
I'd rather not add them to the role, though, because these packages would only be useful in certain use cases (e.g. when developing with python-based software), and not universally for all Docker users.

Issue #346, opened 2022-04-23
Docker installation fails in Arch Linux for Ansible >= 2.12


adamantike:
Related logs:
> TASK [geerlingguy.docker : Install Docker (Ansible >=2.12).] *******************************************************************
> fatal: [host]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (ansible.legacy.pacman) module: allow_downgrade. Supported parameters include: remove_nosave, executable, upgrade, update_cache_extra_args, state, upgrade_extra_args, update_cache (update-cache), extra_args, force, name (package, pkg)."}

Root cause is #336, as the introduced parameter `allow_downgrade` is not supported by the `pacman` manager module.

amotl:
Dear Michael and Jake,

sorry for the annoyance. #353 might have a fix, can I humbly ask you to test it in order to confirm it works before un-drafting the PR?

With kind regards,
Andreas.

RealOrangeOne:
Your patch is worked perfectly for me: https://github.com/geerlingguy/ansible-role-docker/pull/352#issuecomment-1129685797

amotl:
Thank you Jake and sorry for the delay. I've just flagged #352 for review.

amotl:
Dear @adamantike,

@RealOrangeOne confirmed that my fix #352 worked and @geerlingguy was so nice to integrate it and publish a new release. So, I think this issue can be closed.

With kind regards,
Andreas.

adamantike:
Thanks a lot for your quick resolution, @amotl!

I was just able to test the fix, and can also confirm it is working as expected. For anyone else experiencing this issue, version `4.2.3` is the first one to include @amotl's fix.

Glad to close this issue, thanks again!

PR #352, opened 2022-05-16
Fix package downgrading for non-{RedHat,Debian} systems

amotl:
Dear @adamantike,

thank you very much for your report at #346:

> Root cause is #336, as the introduced parameter `allow_downgrade` is not supported by the `pacman` manager module, so that using Ansible >= 2.12 on Arch Linux would croak like:
>
>     Unsupported parameters for (ansible.legacy.pacman) module: allow_downgrade.

I feel responsible for that regression, because I just had focussed on Debian and apparently completely missed the big picture. Apologies.

I am submitting this patch to improve the situation as a draft, because I haven't been able to test it hands-on yet. Do you think it would resolve the issue for you? Apparently, as the software tests seem to succeed, can I humbly ask you to give it a try and report back about the outcome?

In order to pull the patched recipe directly from the git repository, you would put something along those lines into your `requirements.yaml` file.
```yaml
roles:
  - name: ansible-role-docker
    src: git+https://github.com/crate-workbench/ansible-role-docker.git
    version: "amo/allow-downgrades-debian-redhat-only"
```

With kind regards,
Andreas.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -5,22 +5,22 @@
 - include_tasks: setup-Debian.yml
   when: ansible_os_family == 'Debian'
 
-- name: Install Docker (Ansible <2.12).
+- name: Install Docker.
   package:
     name: "{{ docker_package }}"
     state: "{{ docker_package_state }}"
   notify: restart docker
   ignore_errors: "{{ ansible_check_mode }}"
-  when: "ansible_version.full is version_compare('2.12', '<')"
+  when: "ansible_version.full is version_compare('2.12', '<') or ansible_os_family not in ['RedHat', 'Debian']"
 
-- name: Install Docker (Ansible >=2.12).
+- name: Install Docker (with downgrade option).
   package:
     name: "{{ docker_package }}"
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
Tested working correctly with ansible 2.12.5:

```
TASK [geerlingguy.docker : Install Docker (Ansible <2.12).] *********************************************************************************************
ok: [xxxx]

TASK [geerlingguy.docker : Install Docker (Ansible >=2.12).] ********************************************************************************************
skipping: [xxxx]
```

Clearly picks the `<2.12` version

amotl:
Dear Jeff,

I know your time is scarce. Can I still humbly ask you to bring in this patch and cut another release? It will fix #346, so the Arch Linux users from the community, like @adamantike and @RealOrangeOne, will be grateful.

Thank you very much in advance for taking the time and apologies for the hiccup introduced by my recent patch #336.

With kind regards,
Andreas.

amotl:
Thank you very much!

PR #353, opened 2022-05-16
CI: Add Arch Linux to the test matrix

amotl:
Hi again,

#336 introduced a regression because a parameter has been added which is only present for Red Hat and Debian systems, but croaks, for example, on Arch Linux. Thanks for reporting it through #346, @adamantike.

When adding other Linux distributions to the test suite, errors like this might get caught early. If this patch turns out to work sucessfully and is welcome in general, #352 should be based upon it.

With kind regards,
Andreas.

P.S.: This patch has been inspired by suggestions from others. Kudos!
- https://github.com/ansible-community/molecule/discussions/3410
- [How to perform string manipulation while declaring env vars in GitHub Actions](https://stackoverflow.com/a/60445893)
- [Error When Testing Archlinux with Ansible Molecule](https://stackoverflow.com/questions/67172650/error-when-testing-archlinux-with-ansible-molecule)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,10 @@ jobs:
     name: Molecule
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         distro:
+          - archlinux
           - rockylinux8
           - centos7
           - ubuntu2004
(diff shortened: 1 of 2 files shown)


amotl:
This still fails at [1], like:
```
fatal: [instance]: UNREACHABLE! => {"changed": false, "msg": "Failed to create temporary directory.In some cases, you may have been able to authenticate and did not have permissions on the target directory. Consider changing the remote tmp path in ansible.cfg to a path rooted in \"/tmp\", for more error information use -vvv. Failed command was: ( umask 77 && mkdir -p \"` echo ~/.ansible/tmp `\"&& mkdir \"` echo ~/.ansible/tmp/ansible-tmp-1652730294.824226-2282-98062810097350 `\" && echo ansible-tmp-1652730294.824226-2282-98062810097350=\"` echo ~/.ansible/tmp/ansible-tmp-1652730294.824226-2282-98062810097350 `\" ), exited with result 1", "unreachable": true}
```

An extensive report about the problem can be found at https://stackoverflow.com/a/67346654.

However, different kinds of variations of those additonal options in `molecule.yml` did not help.

```yaml
    tmpfs:
      - /run
      - /tmp
    volumes:
      - /sys/fs/cgroup:/sys/fs/cgroup:rw
```

It looks like the molecule container launcher has specific problems on bringing up an Arch Linux instance.

[1] https://github.com/geerlingguy/ansible-role-docker/runs/6458968867#step:7:113

PR #362, opened 2022-06-24
Make service management entirely optional

spawnia:
The provided advice of circumventing service management
did not work for me, usage still failed in WSL:

```
TASK [geerlingguy.docker : Ensure Docker is started and enabled at boot.] **********************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Service is in unknown state", "status": {}}
```

Thus, this adds the option `docker_service_managed`
that allows skipping any service management completely.


geerlingguy:
See: https://github.com/geerlingguy/ansible-role-docker/pull/340

PR #365, opened 2022-07-13
#205 - install Docker packages explicitly

### LINKED ISSUE #205: Install docker-ce-cli & containerd.io
It looks like this role doesn't install `docker-ce-cli` & `containerd.io` packages?
Is there a different role/way to add these?

```
    packages:
    - "docker-ce-{{ docker_version }}"
    - "docker-ce-cli-{{ docker_version }}"
    - containerd.io
```

j-koehler:
Fixes #205.

As [requested here](https://github.com/geerlingguy/ansible-role-docker/issues/205#issuecomment-1068121131) explicitly install all Docker packages and changes the variable(s) name(s).

One note/suggestion: 

You requested to change the name of the variable(s) which makes this a breaking change for the users which configure/override this variable(s). May I ask why? Is it just to be more clear and for naming conventions? Because there is no real need for this. The `package` module supports both strings and lists. 

Keeping it the old way but with a slightly wrong name it would not be a breaking change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,11 +43,11 @@ jobs:
         distro:
           - rockylinux8
           - centos7
+          - ubuntu2204
           - ubuntu2004
           - ubuntu1804
           - debian11
           - debian10
-          - debian9
           - fedora34
 
     steps:
--- a/README.md
+++ b/README.md
@@ -14,10 +14,16 @@ Available variables are listed below, along with default values (see `defaults/m
 
     # Edition can be one of: 'ce' (Community Edition) or 'ee' (Enterprise Edition).
     docker_edition: 'ce'
-    docker_package: "docker-{{ docker_edition }}"
-    docker_package_state: present
+    docker_packages:
+        - "docker-{{ docker_edition }}"
+        - "docker-{{ docker_edition }}-cli"
+        - "docker-{{ docker_edition }}-rootless-extras"
+    docker_packages_state: present
 
-The `docker_edition` should be either `ce` (Community Edition) or `ee` (Enterprise Edition). You can also specify a specific version of Docker to install using the distribution-specific format: Red Hat/CentOS: `docker-{{ docker_edition }}-<VERSION>`; Debian/Ubuntu: `docker-{{ docker_edition }}=<VERSION>`.
+The `docker_edition` should be either `ce` (Community Edition) or `ee` (Enterprise Edition). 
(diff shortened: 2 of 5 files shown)


geerlingguy:
> Is it just to be more clear and for naming conventions?

Yes; with a view towards maintaining this role for 5+ years now, I'd rather take a hit on the naming once and have it make sense a few years down the line, instead of keeping the name `docker_package` and have it default to more than one.

Bumping semver a major version should make people take a glance at the change.

j-koehler:
Alright, I'll have a look at the failing Debian 9 pipeline tomorrow. :)

j-koehler:
> Would you also add `containerd.io` as a package?

I answered the question for myself and added it to the list of packages to be as close as possible to the the official Docker installation instructions.

> Alright, I'll have a look at the failing Debian 9 pipeline tomorrow. :)

I did this and found the reason: the `docker-ce-rootless-extras` package is not available under Debian 9. 
Since we decided to go the major version route I dropped the support for Debian 9 (it's also no longer listed under "OS requirements" [for Debian](https://docs.docker.com/engine/install/debian/)) and while I was on it, I added support for Ubuntu 22.04 LTS (Jammy). Hope you are ok with this. :see_no_evil: 

(Debian 9 should still work ([the files are still there](https://download.docker.com/linux/debian/dists/)), but you would have to overwrite the packages list and figure it out for yourself. Which is (to me) ok for a major version and there is no real need to support it with more complicated defaults and/or tasks with version checks. Except ofc you decide that you still want to support Debian 9.)

j-koehler:
In addition: I also removed `xenial` (Ubuntu 16.04 LTS) from the `meta/main.yml` since there is no longer official support for this version and there are no molecule tests for this (which would probably fail for the same reason as the Debian 9 one). (Force pushed this change.)

geerlingguy:
@j-koehler Heh... well in that case, it's a good thing we're gonna do a major version bump, that'll definitely get people to look twice since there are a few potentially breaking changes here.

j-koehler:
Do you have "release cycles" for your projects? Asking if I use my fork for the time being or if there is a chance there will be a release soon™ :)

geerlingguy:
Thanks for keeping this PR alive—merged and with a couple other changes I've pushed out a 5.0.0 release.

j-koehler:
Thank YOU for providing the first good OSS contribution experience to me. :) My experience on contributions in the past was rather.. "unwelcoming".

Issue #335, opened 2022-03-15
Role fails if docker repository is already on the host but wasn't added by ansible


kaysond:
I'm running on Debian 11. The role failed because I already had an existing file in `/etc/apt/sources.list.d/` called `docker.list`. This is how the Docker install instructions recommend adding it when installing manually: https://docs.docker.com/engine/install/debian/#set-up-the-repository

I took a quick look and am guessing this is just a limitation of `ansible.builtin.apt_repository`, which adds a file called `download_docker_com_linux_debian.list` with the same repo. I guess it's not smart enough to actually check what repos are installed first, because its not exactly trivial.

As a workaround, it might be worth adding a task to grep the files in `/etc/apt/sources.list.d` for the repository that is about to be added and skip if it already exists. Or just delete any `/etc/apt/sources.list.d/docker.list`. Or something similar...

```
E: Conflicting values set for option Signed-By regarding source https://download.docker.com/linux/debian/ bullseye: /usr/share/keyrings/docker-archive-keyring.gpg !=
E: The list of sources could not be read.
```

kaysond:
> this role is meant to be used on a clean installation where Docker has not been set up before
Is there something in particular about the role that makes this so? It should be idempotent, right? Aside from this particular issue with the `apt_repository` module, I didn't see anything else that would cause problems. And indeed, when I removed `docker.list` from my host that had docker installed already, everything worked fine. It also upgrades the docker version just fine.

kaysond:
To be fair, it may be better to fix this upstream...

stucki:
I just came here because I wanted to request changing the apt filename from `download_docker_com_linux_debian.list` to simply `docker.list` as explained in the upstream Docker installation tutorial (https://docs.docker.com/engine/install/debian/).

Is there a reason for not setting the `filename` when calling the `apt_repository` module? I suggest to set it to `docker` just as explained in the upstream tutorial.

It would also be possible to make this configurable as proposed in #101, however this PR has been closed already.

kaysond:
Not stale

geerlingguy:
I'd be willing to merge a PR that makes the filename configurable — that issue (#101) was an issue, not a PR though?

geerlingguy:
[...]

Issue #101, opened 2018-12-01
Add a variable docker_apt_filename to control apt repository filename


julien-lecomte:
Feature Request:

Add a variable docker_apt_filename to control apt repository filename. Default to ansible apt_repository default's (https://docs.ansible.com/ansible/2.6/modules/apt_repository_module.html)

geerlingguy:
Can you give a little more detail as to why you would like more control over the filename? I haven't had to do this before, but if it's a common need I would be okay supporting it.

julien-lecomte:
Is OCD a valid reason? ;-)

My reason is that I have legacy scripts that use the file names to activate / deactive sources and packages and other actions based on filename. ie: On the buster system I have under my hand right now I have debian-buster.list, debian-buster-updates.list, debian-buster-security.list, and docker-ce.list; each with only one line.

I don't know how common this is, but at worst it adds an optional functionality without breaking existing behavior.
I'd reason that if it was added to ansible in 2.1, there must have been a need.

julien-lecomte:
Hi
May I ask for an update on this? Most notably, will you accept to merge the associated MR?
Thanks,
Julien

PR #368, opened 2022-08-09
Add filename option to apt repository add

kaysond:
Addresses #335 and #101 (i.e. I'd say it's technically a work around as opposed to a solution)

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ The main Docker repo URL, common between Debian and RHEL systems.
     docker_apt_repository: "deb [arch={{ docker_apt_arch }}] {{ docker_repo_url }}/{{ ansible_distribution | lower }} {{ ansible_distribution_release }} {{ docker_apt_release_channel }}"
     docker_apt_ignore_key_error: True
     docker_apt_gpg_key: "{{ docker_repo_url }}/{{ ansible_distribution | lower }}/gpg"
+    docker_apt_filename: ""
 
 (Used only for Debian/Ubuntu.) You can switch the channel to `nightly` if you want to use the Nightly release.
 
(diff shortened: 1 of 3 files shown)


kaysond:
Not stale

kaysond:
not stale (unless superseded by #369)

README.md
```diff
@@ -56,6 +56,7 @@ The main Docker repo URL, common between Debian and RHEL systems.
     docker_apt_repository: "deb [arch={{ docker_apt_arch }}] {{ docker_repo_url }}/{{ ansible_distribution | lower }} {{ ansible_distribution_release }} {{ docker_apt_release_channel }}"
     docker_apt_ignore_key_error: True
     docker_apt_gpg_key: "{{ docker_repo_url }}/{{ ansible_distribution | lower }}/gpg"
+    docker_apt_filename: omit
```
geerlingguy, line 59: For this option, could you just put instead of `omit`, `custom-filename-here-if-desired` or something that would be a more practical example. Or maybe we could leave it empty in `defaults/main.yml` instead, that might be preferable (`docker_apt_filename: ''`, and then document that value here, so it is more obvious it should be a string.

geerlingguy (changes requested):

README.md
kaysond: Agree that the latter is probably preferable. Unfortunately, it seems like you can't give an empty string for that argument. This PR fail CI, but #369, which has a non-empty string, passes CI.

We could hack around it with something like `filename: {{ docker_apt_filename if docker_apt_filename != "" else omit }}`, assuming that works.

Alternatively, it may make more sense to just close this in favor of #369, if you plan on releasing 7.0.0 soon. This PR is mainly intended as a temporary fix until the defaults get updated anyway.

kaysond:
not stale

dsegurag:
@geerlingguy, @kaysond what if we add something like `filename: "{{ docker_apt_filename | default(omit) }}"`, and we don't need to set anything under "default" folder:

```yaml
- name: Add Docker repository.
  apt_repository:
    repo: "{{ docker_apt_repository }}"
    filename: "{{ docker_apt_filename | default(omit) }}"
    state: present
    update_cache: true
  when: docker_add_repo | bool
```

It's backward compatible with the current code. For next major version we can set a name (maybe "docker") for the file, but meanwhile people can start changing it to the desired name.

kaysond:
Superseded by #369

PR #369, opened 2022-08-09
Change defaults in the next major release

kaysond:
Setting up this PR to update some default options the next time a major release is going to happen:

- apt repository addition in Debian will by default use the file `docker.list` which matches the docker documentation, and solves #335 
- install the `docker compose` plugin via the docker repo package instead of the docker-compose standalone binary by default

For the latter


geerlingguy:
Has a couple conflicts now, but marking this as planned since I want to do it soon-ish.

kaysond:
Rebased

kaysond:
@geerlingguy any updates on when you'll do the next major release?

kaysond:
Bump. @geerlingguy - can we get a 7.0.0?

geerlingguy:
There's currently a merge conflict in this PR but keep bumping once that's fixed and we'll get there :)

kaysond:
@geerlingguy done

gizero:
@kaysond @geerlingguy Having this merged caused me the same problem it was supposed to fix (#335) but the other way around. I had docker installed by a previous version of this role and upgrading the role to 7.0.0 led me into the situation where two different apt sources where present and conflicting:

```
fatal: [slapp01]: FAILED! => {"changed": false, "msg": "E:Conflicting values set for option Signed-By regarding source https://download.docker.com/linux/ubuntu/ jammy: /etc/apt/trusted.gpg.d/docker.asc != , E:The list of sources could not be read."}
```

I understand this change to the role's defaults was released as a major version bump which may introduce breaking changes, but I don't clearly see what is the right upgrade path supposed to be. Should I have added in advance a task to my playbooks to ensure the existing apt source file (the one which name is automatically generated by the `apt_repository` module when no `filename` option was given) is removed if present? If that's the case, I don't see the real point of this change. Could you please help understanding? TIA

kaysond:
> @kaysond @geerlingguy Having this merged caused me the same problem it was supposed to fix (#335) but the other way around. I had docker installed by a previous version of this role and upgrading the role to 7.0.0 led me into the situation where two different apt sources where present and conflicting:
> 
> ```
> fatal: [slapp01]: FAILED! => {"changed": false, "msg": "E:Conflicting values set for option Signed-By regarding source https://download.docker.com/linux/ubuntu/ jammy: /etc/apt/trusted.gpg.d/docker.asc != , E:The list of sources could not be read."}
> ```
> 
> I understand this change to the role's defaults was released as a major version bump which may introduce breaking changes, but I don't clearly see what is the right upgrade path supposed to be. Should I have added in advance a task to my playbooks to ensure the existing apt source file (the one which name is automatically generated by the `apt_repository` module when no `filename` option was given) is removed if present? If that's the case, I don't see the real point of this change. Could you please help understanding? TIA

@gizero - you just needed to set the `docker_apt_filename` variable to whatever the filename is on your existing installs. For example, on debian, the default was `download_docker_com_linux_debian` (the module adds `.list` automatically). You probably need to delete one of the files if you've currently got both.

The reason for the change is to align with the [Docker installation docs](https://docs.docker.com/engine/install/debian/#install-using-the-repository).

This could probably be more clear in the docs... There might be a way to detect the old one and set the variable dynamically, but it might need to be hardcoded by OS

Issue #308, opened 2021-10-21
Some kind of notification post group add


impala454:
Hi Jeff, long time subscriber to your channel, and very much appreciate this role.  I was thinking of submitting a PR for this but curious on your thoughts first.  I use this role to ensure docker is installed on a machine before then pulling some images.  On a fresh build I add my user to the `docker_users` variable and that properly gets set, but if that got changed I need some kind of notification to kick the ssh session or whatever so the user can then actually use docker.  I tried but can't see a way to get such notification without modifying this role (which I happily will let you maintain this role).  Does it make sense to add some kind of notify to this role or is there some way I'm missing to see it changed and take action myself in a task?

geerlingguy:
I can't remember why exactly, but I think I had been testing it and for some reason at least with Docker, that wasn't working for me (or maybe a particular use case for the role)...

geerlingguy:
Heh... oops? https://github.com/geerlingguy/ansible-role-docker/issues/237#issuecomment-832760600

impala454:
I was about to go ahead and PR this but realized in your other closed issue #237 you requested the connection get reset only if user or groups changed.  It appears this is not possible (see [here](https://stackoverflow.com/a/44753457/1544725)).  So I suppose we could either add it to always happen or just skip it altogether.  Up to you.

impala454:
I just ended up adding it to my play after the role call and before my tasks that use docker.

Issue #237, opened 2020-11-01
Reboot required when when using docker_users feature


domenicbove:
I am observing on ubuntu and centos amis, that after running your role:

```
- name: Prepare
  hosts: all
  roles:
    - role: geerlingguy.docker
      vars:
        docker_users:
          - "{{ansible_user}}"
      become: yes
```

A later playbook which has `/usr/local/bin/docker-compose up -d` without `become:true` fails with:
```
Couldn't connect to Docker daemon at http+docker://localhost - is it running?\n\nIf it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.
```

This is all resolved by doing:
```
- name: Prepare
  hosts: all
  roles:
    - role: geerlingguy.docker
      vars:
        docker_users:
          - "{{ansible_user}}"
      become: true

- name: Prepare
  hosts: all
  tasks:
    - name: Reboot of host so ansible_user will work with Docker
      reboot:
      become: true

    - name: Wait for SSH
      wait_for:
        port: 22
        host: "{{ ansible_host }}"
        search_regex: SSH
        delay: 10
        timeout: 320
      connection: local
```

Do you think the role should be handling the reboot?

fourstepper:
I don't think reboot is necessary, but the playbook that is run afterwards might be using the same SSH (thus login) session connection

Could you try waiting between the runs to see if this is the case? Thanks!

domenicbove:
@fourstepper What would be a good way to "wait between runs"

I am automating a provisioning playbook that (this is an oversimplification):
- Creates vms
- Uses this role to install docker + adds the ansible_user to the docker group
- Uses the ansible_user to create containers

Imagine this is all in one "playbook" or in multiple playbooks strung together, I found the reboot was necessary, but I agree I'd like to move away from it if possible

fourstepper:
I havent tried this, but it should do the work

https://serverfault.com/questions/930804/force-ansible-to-log-off-to-refresh-user-groups

domenicbove:
Wow awesome that looks like what I need. 

For the sake of conversation, should `meta: reset_connection`
- be added to this role?
- or be added to the docs?

fourstepper:
I am not to answer that question - that's more up to @geerlingguy

I am sure he would happily review either of the options a pull requests though and carry the discussion over there :)

geerlingguy:
[...]

PR #370, opened 2022-08-24
reset ssh connection to allow user changes to affect current login user

fardke:
### Linked issues:
* https://github.com/geerlingguy/ansible-role-docker/issues/237
* https://github.com/geerlingguy/ansible-role-docker/issues/308

### Why:
Without this modification, we can't use docker in same playbook than the one which install docker (in the case of we use `docker_users` variable).

### How:
Check that `docker_users` are set and not in docker group before include `docker-users.yml`. In `docker-users.yml` we call reset_connection` from [ansible.builtin.meta](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/meta_module.html) collection after `docker_users` are added to docker group.

### Manual success tests:
* Try to install docker with only one user in docker-users and not in docker group => docker-users.yml include
* Try to install docker with only one user in docker-users but the user is in docker group => docker-users.yml not include
* Try to install docker with 2 users in docker-users, one user in docker group and the second not => docker-users.yml include
* Try to install docker with 2 users in docker-users, both are in docker group => docker-users.yml not include
* Try to install docker with 2 users in docker-users, both are not in docker group => docker-users.yml include

### DIFF
--- a/tasks/docker-users.yml
+++ b/tasks/docker-users.yml
@@ -5,3 +5,7 @@
     groups: docker
     append: true
   with_items: "{{ docker_users }}"
+
+- name: Reset ssh connection to apply users changes
+  meta:
+    reset_connection
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -79,5 +79,20 @@
 - include_tasks: docker-compose.yml
   when: docker_install_compose | bool
 
-- include_tasks: docker-users.yml
+- name: Get all docker group infos
+  getent:
+    database: group
+    key: docker
+    split: ':'
   when: docker_users | length > 0
+
+- name: Check there is at least one user to add to docker group
+  set_fact:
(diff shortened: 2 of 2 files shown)


geerlingguy:
I had tried using that `meta` trick before but never could get it to pick things up in a playbook run over SSH correctly... maybe it was a bug in Ansible that's fixed now?

fardke:
Hello,
I never try to use `meta reset_connection` before, but on ansible core 2.13.3 it works directly. 
Some issues are closed on ansible related to this :
- https://github.com/ansible/ansible/issues/50306
- https://github.com/ansible/ansible/issues/66414

Two issues seems to be fixed.

geerlingguy:
Heh, I even wrote an article on this: https://www.jeffgeerling.com/comment/31175#comment-31175

Issue #372, opened 2022-09-05
Deprecated GPG usage


sosheskaz:
Same issue as here: https://github.com/geerlingguy/ansible-role-jenkins/pull/355/commits/94e6270a60fd931788e187952bc24bc532692b1a

PR #373, opened 2022-09-05
Download apt key directly

sosheskaz:
Fixes issue #372


sosheskaz:
I have tested this on my Ubuntu Jammy machine, but nothing else yet.

PR #381, opened 2022-10-26
Upgrade the OpenPGP key retrieval

ictus4u:
As the apt-key method has been deprecated, this commit includes the recommended `signed-by` directive in the repository definition.

Followed the tips here:
https://www.linuxuprising.com/2021/01/apt-key-is-deprecated-how-to-add.html

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -31,10 +31,11 @@ docker_repo_url: https://download.docker.com/linux
 
 # Used only for Debian/Ubuntu. Switch 'stable' to 'nightly' if needed.
 docker_apt_release_channel: stable
-docker_apt_arch: "{{ 'arm64' if ansible_architecture == 'aarch64' else 'amd64' }}"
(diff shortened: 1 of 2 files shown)


geerlingguy:
Please see: https://www.jeffgeerling.com/blog/2022/aptkey-deprecated-debianubuntu-how-fix-ansible

ictus4u:
Oh, I see. Thank you for pointing out the ongoing discussion.
I've seen the `--dearmor` approach is under debate. However, please consider that without the `signed-by` directive [pointing to a separate non-widely trusted file](https://itsfoss.com/apt-key-deprecated/#:~:text=You%20should%20edit,more%20cross%2Dsigning.), we keep the cause for the deprecation in the first place: the risk of cross-signing.
The https://github.com/tonyclemmey/ansible-role-mongodb/commit/ca676ed235f21de8041ec2f99baeadc02e08248b solution goes in the way of avoiding to `--dearmor` but is honoring the `signed-by` requirement.
If you find it worth it, I could make another PR.

PR #394, opened 2022-12-14
Add flag to deactivate repository configuration

aurelien-bareille:
When using on environments with already pre-configured repositories it is useful to be able to skip tasks configuring a specific docker repository

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -26,6 +26,9 @@ docker_compose_arch: "{{ ansible_architecture }}"
 docker_compose_url: "https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-linux-{{ docker_compose_arch }}"
 docker_compose_path: /usr/local/bin/docker-compose
 
+# Enable repo setup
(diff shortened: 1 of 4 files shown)


defaults/main.yml
```diff
@@ -26,6 +26,9 @@ docker_compose_arch: "{{ ansible_architecture }}"
 docker_compose_url: "https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-linux-{{ docker_compose_arch }}"
 docker_compose_path: /usr/local/bin/docker-compose
 
+# Enable repo setup
+docker_add_repo: true
```
geerlingguy, line 30: Can you add the variable to the README.md file with a description underneath like "Controls whether this role will add the official Docker repository. Set to `false` if you want to use the default docker packages for your system or manage the package repository on your own.

geerlingguy (changes requested):

defaults/main.yml
aurelien-bareille: Doc updated

PR #395, opened 2022-12-16
Add a variable to prevent auto-start on package installation

lel-amri:
At the time of committing :
* No RPM packages for CentOS has an action that start the service in post-installation scripts
* All DPKG packages for Debian, Rasbian and Ubuntu have an action that start the service in post-installation scripts

This commit adds the `docker_service_prevent_autostart` variable. When set to `true`, prevents service start on package installation.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -9,6 +9,7 @@ docker_packages:
 docker_packages_state: present
 
 # Service options.
+docker_service_prevent_autostart: false
 docker_service_manage: true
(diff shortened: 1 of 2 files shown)


lel-amri:
This is my first contribution to an Ansible role. This might not be the way to do things with `ansible.builtin.package`. Commit message might have grammar errors.
I'm open to any comment regarding this PR as a whole.

lel-amri:
Bump

PR #399, opened 2023-02-04
fix ansible-lint

juju4:


molecule/default/converge.yml
```diff
@@ -5,11 +5,13 @@
 
   pre_tasks:
     - name: Update apt cache.
-      apt: update_cache=yes cache_valid_time=600
+      ansible.builtin.apt:
+        update_cache: yes
```
oDn, line 9: Might as well set update_cache to "true" for _thruthyness_ as elsewhere in the role.

oDn (changes requested):
One minor _thruthyness_ issue to solve to get rid of one remaining ansible-lint warning.

geerlingguy:
Thanks for the suggestions, however I use my own ansible-lint rules to bypass some that I feel are too strict or that I'd rather deal with at a later time... https://github.com/geerlingguy/ansible-role-docker/blob/master/.ansible-lint

PR #403, opened 2023-02-19
added support for Rocky Linux

theko2fi:
Hi there, 
I added the support for Rocky Linux. Please @geerlingguy check it out. This PR is just one line.

Thanks !


geerlingguy:
The OS family should be reported as `RedHat` I believe... see https://github.com/ansible/ansible/pull/74530

PR #406, opened 2023-03-07
Update APT cache on first invocation

abalage:
Explicitly call package cache update only on first occurrence.

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -12,6 +12,7 @@
       - apt-transport-https
       - ca-certificates
     state: present
+    update_cache: true
   when: docker_add_repo | bool
 
 - name: Ensure additional dependencies are installed (on Ubuntu < 20.04 and any other systems).


abalage:
Fixes #407

abalage:
Okay I got the things written in that blog, no problem with that.
However the fix is just a single line, without breaking anything but solving reported issues.
Imho it would be pity not incorporating that. :)

geerlingguy:
This will break idempotence—for my roles, I always manage apt caches at the play level, not at the role level. Otherwise every role I have, I would need to add in `update_cache` and manage a lifetime for the cache, which I'd rather not do since everyone has a different approach. See: https://github.com/geerlingguy/ansible-role-docker/blob/master/molecule/default/converge.yml#L6-L9

abalage:
Yes I went down that path too because of the change was not merged when I made the PR.
Thanks for the feedback though.

PR #413, opened 2023-04-11
Rootless docker

### LINKED ISSUE #412: Option to install rootless
Hi!

I saw that we're installing the Docker rootless package but we're not actually setting it up. In case I just missed that option, please let me know :-)

Here's what I'm doing currently as a workaround:

```yaml
- name: 'Disable root docker.service'
  ansible.builtin.systemd:
    name: docker.service
    enabled: false
  when: docker_rootless and docker_disable_root_daemon == true

- name: 'Disable root docker.socket'
  ansible.builtin.systemd:
    name: docker.socket
    enabled: false
  when: docker_rootless and docker_disable_root_daemon == true

- name: 'Install uidmap for rootless UID mapping'
  ansible.builtin.package:
    name: uidmap
    state: present
  when: docker_rootless

- name: 'Install Docker rootless'
  become: false
  ansible.builtin.shell: dockerd-rootless-setuptool.sh install
  args:
    creates: /home/{{ ansible_user }}/.config/systemd/user/docker.service
  when: docker_rootless

- name: 'Set capabilities for rootlesskit'
  community.general.capabilities:
    path: /usr/bin/rootlesskit
    capability: '{{ docker_capabilities | join(",") }}=ep'
    state: present
  register: docker_capabilities_update
  when: docker_rootless

- name: 'Restart docker service'
  become: false
  ansible.builtin.systemd:
    state: restarted
    name: docker
    scope: user
  when: docker_rootless and docker_capabilities_update.changed

- name: 'Get current user ID'
  getent:
    database: passwd
    key: '{{ ansible_user }}'
  when: docker_rootless

- name: 'Add export for path'
  become: false
  ansible.builtin.lineinfile:
    path: /home/{{ ansible_user }}/.bashrc
    line: export path=/usr/bin:$PATH
    create: yes
  when: docker_rootless

- name: 'Add export for DOCKER_HOST'
  become: false
  ansible.builtin.lineinfile:
    path: /home/{{ ansible_user }}/.bashrc
    line: export DOCKER_HOST=unix:///run/user/{{ getent_passwd[ansible_user][1] }}/docker.sock
    create: yes
  when: docker_rootless
```

with the following variables:

```yaml
docker_rootless: true
docker_disable_root_daemon: true
docker_capabilities:
  - cap_net_bind_service
```

Is this something that could find its way into the main package?

reboss:
Here is a fairly complete "Rootless Docker" setup. 

One can toggle between both root Docker and rootless Docker by re-running the role with `docker_rootless` set to either `true` or `false`

I maintain idempotency when running `command` tags so that we don't run the `docker-rootless-setuptool.sh < install | uninstall >` unnecessarily. 

There are a couple things that I didn't get to, and don't see much value in solving for this PR:
1. This code doesn't support rootless docker in RHEL/CentOS 7 because both are end of life and I feel like trying to get them to work may overcomplicate the code a bit.
2. I chose not to install fuse-overlayfs, or any of its counterparts, despite it being recommended by the docs.  Once again, I feel like this would have overcomplicated things a bit. It's only recommended Docker, and not a strict dependency anyways.

Ultimately, I have tested this thoroughly on Ubuntu and it should work on the latest version of RHEL, however I can't test because I don't have a RHEL subscription.

I'd say having this support one OS, is better than supporting none.  Let me know if you'd like any changes made, @geerlingguy 

Fixes #412 

Update: I've confirmed that this process works on CentOS Stream 9

### DIFF
--- a/tasks/docker-rootless.yml
+++ b/tasks/docker-rootless.yml
@@ -26,12 +26,13 @@
 - name: Install rootless docker
   become: false
   command: /usr/bin/dockerd-rootless-setuptool.sh install
+  when: rootless_conf.stat.exists == false
 
 - name: Enable and start rootless docker
   become: false
   systemd:
-    name: docker
-    state: started
+    name: docker.service
+    state: restarted
     enabled: true
     scope: user
 
--- a/tasks/docker-rootless.yml
+++ b/tasks/docker-rootless.yml
@@ -1,0 +1,45 @@
+---
+- name: Ensure dockerd-rootless-setup.sh is installed
+  apt:
+    name:
+      - uidmap
+      - docker-ce-rootless-extras
+    state: present
+
+- name: Stop any running root instances of docker daemon
+  systemd:
+    name: docker.service
+    state: stopped
+    enabled: false
+
+- name: Close root docker socket
+  systemd:
+    name: docker.socket
+    state: stopped
+    enabled: false
+
+- name: Remove docker.sock file
+  file:
+    path: /var/run/docker.sock
+    state: absent
+
+- name: Install rootless docker
+  become: false
+  command: /usr/bin/dockerd-rootless-setuptool.sh install
+
+- name: Enable and start rootless docker
+  become: false
+  systemd:
+    name: docker
+    state: started
+    enabled: yes
+    scope: user
+
+- name: Decouple rootless docker from user session
+  command: loginctl enable-linger {{ ansible_user }}
+
+- name: Add DOCKER_HOST to systemwide environment file
+  lineinfile:
+    path: /etc/environment
+    insertafter: EOF
+    line: 'DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock'
--- a/tasks/docker-rootless.yml
+++ b/tasks/docker-rootless.yml
@@ -36,10 +36,10 @@
     scope: user
 
 - name: Decouple rootless docker from user session
-  command: loginctl enable-linger {{ ansible_user }}
+  command: "loginctl enable-linger {{ ansible_user }}"
 
 - name: Add DOCKER_HOST to systemwide environment file
   lineinfile:
     path: /etc/environment
     insertafter: EOF
-    line: 'DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock'
+    line: "DOCKER_HOST=unix://{{ lookup('env', 'XDG_RUNTIME_DIR') }}/docker.sock"
--- a/tasks/docker-rootless.yml
+++ b/tasks/docker-rootless.yml
@@ -32,7 +32,7 @@
   systemd:
     name: docker
     state: started
-    enabled: yes
+    enabled: true
     scope: user
 
 - name: Decouple rootless docker from user session
--- a/tasks/docker-rootless.yml
+++ b/tasks/docker-rootless.yml
@@ -1,10 +1,19 @@
 ---
 - name: Ensure dockerd-rootless-setup.sh is installed
-  apt:
+  package:
     name:
       - uidmap
       - docker-ce-rootless-extras
     state: present
+  when: ansible_distribution != "CentOS"
+
+- name: Ensure dockerd-rootless-setup.sh is installed
+  package:
+    name:
+      - shadow-utils
+      - docker-ce-rootless-extras
+    state: present
+  when: ansible_distribution == "CentOS"
 
 - name: Stop any running root instances of docker daemon
   service:
@@ -22,6 +31,10 @@
   file:
     path: /var/run/docker.sock
     state: absent
+
+- name: Modprobe ip_tables
+  modprobe:
+    name: ip_tables
 
 - name: Install rootless docker
   become: false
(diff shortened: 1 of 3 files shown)


tasks/docker-rootless.yml
```diff
@@ -0,0 +1,60 @@
...
+
+- name: Ensure dockerd-rootless-setup.sh is installed
+  package:
+    name:
+      - shadow-utils
+      - docker-ce-rootless-extras
+    state: present
+  when: ansible_distribution == "CentOS"
+
+- name: Stop any running root instances of docker daemon
```
paranerd, line 18: It may just be me but going completely rootless caused quite a lot of issues with permissions, mapping multiple users/uids (like root AND www-data for example), etc. So I would like stopping/removing the root docker daemon to be optional to be able to run Docker in a "dual setup".

tasks/setup-Debian.yml
```diff
@@ -6,6 +6,30 @@
...
+
+- name: Add kernel.unprivileged_userns_clone=1 to /etc/sysctl.conf
+  lineinfile:
+    path: /etc/sysctl.conf
+    insertafter: EOF
+    line: 'kernel.unprivileged_userns_clone=1'
+  when:
+    - docker_rootless == true
+    - ansible_distribution == 'Debian'
+    - ansible_distribution_version == '10'
```
geerlingguy, line 31: Would this also apply to later versions of Debian, or *only* 10?

reboss: This only applies to Debian 10.  The [docs](https://docs.docker.com/engine/security/rootless/) say:

> For Debian 10, add kernel.unprivileged_userns_clone=1 to /etc/sysctl.conf (or /etc/sysctl.d) and run sudo sysctl --system. This step is not required on Debian 11.

Although, technically I haven't tested on a pure Debian machine yet, just Ubuntu.  I'll try to test it today but I hope I don't hit another subscription wall like I did with RHEL 😅

tasks/docker-rootless.yml
reboss: This is a good idea, I'd be happy to try it but I didn't think it was possible.  When I run the `dockerd-rootless-setuptool.sh` script to install, I get an error if the docker.sock file exists, even if it isn't listening, which is why I have to explicitly remove it in the following step.  

There is the `--force` option, but I don't know how safe that is to use.

It looks like that's what your script does in #412.  What flavor of Linux did you use that script on?

paranerd: I ran it under Ubuntu Server 22.04 LTS. Just ran it again and can confirm that it indeed doesn't work. Probably added the condition after the fact, sorry for the confusion 😁

reboss: No problem.  It's too bad it doesn't work, it would be nice to have an option.  Not sure how it could be enforced though - I believe users still need to be in the docker group to run rootless docker, so if root docker is still available to them, it kind of defeats the purpose of locking down the host IMO.

paranerd: I was able to run rootless Docker without my user being in the "docker" group.

tasks/setup-Debian.yml
reboss: Ok so I tested this on Debian 10 and it looks like the docs are a bit misleading.  It says `slirp4netns >= 0.4.0` is required, however, that version doesn't seem to be available using the package manager.

I tested it on Debian 11 though and it works just fine, so I've removed the extra setup needed for Debian 10 altogether.

So all-in-all, it supports all of the latest versions, Ubuntu 20 and above, CentOS Stream 9, and Debian 11.

I could update the docs and state that this role only supports rootless docker for the latest versions, and maybe put a disclaimer that says, "rootless docker is still experimental, and if you need to revert back to a normal installation, you can just run the role again with `docker_rootless` set to false".

I have tested toggling between the two pretty thoroughly and haven't encountered any errors.

Let me know what you think!  Thanks :)

tasks/docker-rootless.yml
lel-amri, line 18: You might be interested by what I did in #395.

tasks/docker-rootless.yml
```diff
@@ -0,0 +1,60 @@
...
+  ignore_errors: "{{ ansible_check_mode }}"
+
+- name: Decouple rootless docker from user session
+  command: "loginctl enable-linger {{ ansible_user }}"
+
+- name: Add DOCKER_HOST to systemwide environment file
+  lineinfile:
+    path: /etc/environment
+    insertafter: EOF
+    line: "DOCKER_HOST=unix://{{ lookup('env', 'XDG_RUNTIME_DIR') }}/docker.sock"
```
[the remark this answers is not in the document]
egdoc: What if multiple users want to use docker in rootless mode? /etc/environment is a global file, this would put the XDG_RUNTIME_DIR of that specific user in it. Wouldn't be better to define the "DOCKER_HOST" variable in each user ~/.bash_profile or ~/.bashrc?

joshavant:
Just commenting that I'd be excited for this to get merged into main.

Is there an easy answer to what's left on this and what it needs to get merged into main? @geerlingguy @paranerd @egdoc

dometto:
Reiterating @joshavant's query -- this would be a really great addition to the role, any chance it can be revived?

PR #418, opened 2023-06-15
APT key: don't force and provide checksum, to fix check mode

nevart:
As described in https://github.com/ansible/ansible/issues/65687, get_url only partially supports check_mode: "the changed status will reflect comparison to an empty source file".

Before this change, executing this code, with the key already being in place on the target system, would report "OK", while check_mode would report "changed".

Due to this change, both now either report "OK" or "changed", depending on the state of the target system.

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -31,7 +31,8 @@
     url: "{{ docker_apt_gpg_key }}"
     dest: /etc/apt/trusted.gpg.d/docker.asc
     mode: '0644'
-    force: true
+    force: false
+    checksum: "{{ docker_apt_gpg_key_checksum | default(omit) }}"
   register: add_repository_key
   ignore_errors: "{{ docker_apt_ignore_key_error }}"
   when: docker_add_repo | bool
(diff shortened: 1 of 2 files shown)


tasks/setup-Debian.yml
```diff
@@ -31,7 +31,8 @@
     url: "{{ docker_apt_gpg_key }}"
     dest: /etc/apt/trusted.gpg.d/docker.asc
     mode: '0644'
-    force: true
+    force: false
+    checksum: sha256:1500c1f56fa9e26b9b8f42452a553675796ade0807cdce11975eb98170b3a570
```
geerlingguy, line 35: The issue I see with this approach is that the `checksum` value here would be associated with `docker_apt_gpg_key`, and for maintenance reasons, both would need changes over time (maybe) depending on how Docker maintains their repo. Also, this change could break existing installs for users who don't use the same apt repo.

I think checksum would need to be configurable, at minimum, and the value here could be the default.

geerlingguy (changes requested):

tasks/setup-Debian.yml
geor-g: Thanks for the review and feedback.

> I think checksum would need to be configurable, at minimum, and the value here could be the default.

You mean something like:

```
    checksum: "{{ docker_apt_gpg_key_checksum | default(sha256:1500c1f56fa9e26b9b8f42452a553675796ade0807cdce11975eb98170b3a570) }}"
```

and the people would be able to overwrite it, if required in their environments?

geor-g: @geerlingguy Thanks again for your feedback -- any comment on the above?

geerlingguy: Yes, exactly!

Or just set `checksum: "{{ docker_apt_gpg_key_checksum | default(omit) }}`

and then in the defaults/main.yml vars file:

```
docker_apt_gpg_key_checksum: "sha256:1500c1f56fa9e26b9b8f42452a553675796ade0807cdce11975eb98170b3a570"
```

geerlingguy (approved):

PR #424, opened 2023-09-26
remove runc, as it conflicts with containerd

rpetti:
On RedHat operating systems, runc is provided by default on some installations. This other container implementation conflicts with containerd, making installation impossible without removing it first.

This PR ensures runc is not installed prior to installing containerd so it doesn't fail due to this conflict.


geerlingguy:
This seems to be failing CI

rpetti:
> This seems to be failing CI

Got it, thanks!

PR #430, opened 2023-10-15
Change defaults for docker_apt_arch and docker_compose_arch to support armv7.

DecWebb:
Add conversion for armv7l to armv7 for compatibility with docker compose download architectures.
Replaces hardcoded "amd64" with ansible_architecture to support other architectures than amd64 and arm64.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -37,7 +37,7 @@ docker_apt_release_channel: stable
 # docker_apt_ansible_distribution is a workaround for Ubuntu variants which can't be identified as such by Ansible,
 # and is only necessary until Docker officially supports them.
 docker_apt_ansible_distribution: "{{ 'ubuntu' if ansible_distribution in ['Pop!_OS', 'Linux Mint'] else ansible_distribution }}"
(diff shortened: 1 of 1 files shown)


DecWebb:
I have fixed the build errors resulting from my change. I think some manual tests on different architectures to double check this change would be worthwhile.

geerlingguy:
Is this for older Raspberry Pi installations? I pause when thinking of supporting non-64-bit OSes as Raspberry Pi themselves have finally moved on to 64-bit default for the Pi 5 (and 64-bit works on almost all Pi models now).

DecWebb:
Yes, personally this is for an old rpi2b and also a Helios4 which is an armbian board. Both are 32bit and wont get 64bit OS support. I think its worthwhile while docker still supports 32bit os with the armhf binary (https://download.docker.com/linux/debian/dists/bookworm/stable/).

DecWebb:
Closing as will not get merged, but useful for reference for others with older raspberry pi's and arm boards who run into errors using this role.

Issue #435, opened 2023-11-17
GPG keys should not be saved to `/apt/trusted.gpg.d` / Do not use `apt-key`


kawadeomkar:
As per https://wiki.debian.org/DebianRepository/UseThirdParty, third party keys should not be placed in `/etc/apt/trusted.gpg.d`.
From the Wiki:
```
If future updates to the certificate will be managed by an apt/dpkg package as recommended below, then it SHOULD be downloaded into /usr/share/keyrings using the same filename that will be provided by the package. If it will be managed locally , it SHOULD be downloaded into /etc/apt/keyrings instead.
```

I noticed that the docker docs installation guide https://docs.docker.com/engine/install/debian/ places the docker GPG keys 
in `/etc/apt/keyrings`. 

Also, `apt-key` is deprecated and will not be available after Debian 11 / Ubuntu 22.04. 
Instead of  `apt-key add -` https://github.com/geerlingguy/ansible-role-docker/blob/8ff4a241477f5f24c718d606a2ee450b370cc47c/tasks/setup-Debian.yml#L47
this should be `sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg`

I've started working on the proposed changes, and will make a PR shortly

bryion:
This worked for me.

Issue #434, opened 2023-11-13
Failure when adding docker repository


pantelis-karamolegkos:
I am installing the role as follows

```yaml
    - name: Install required ansible-galaxy roles
      local_action: shell ansible-galaxy install -r requirements.yaml --force
```

where

```
$ cat requirements.yaml
- src: geerlingguy.docker
```

and then invoking it

```yaml
    - name: Install docker and docker-compose
      include_role:
        name: geerlingguy.docker
        apply:
          become: yes
      vars:
        docker_compose_path: "/usr/local/bin/docker-compose"
        docker_compose_version: "1.27.4"
```

This fails with the following error

```
│ TASK [geerlingguy.docker : Add Docker repository.]
│ *****************************
│ An exception occurred during task execution. To see the full traceback, use
│ -vvv. The error was: apt_pkg.Error: E:Conflicting values set for option
│ Signed-By regarding source https://download.docker.com/linux/ubuntu/ focal:
│ /etc/apt/trusted.gpg.d/docker.asc != , E:The list of sources could not be
│ read.
```

`ansible` runs on the following machine

```
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.2 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

Any suggestions?

pantelis-karamolegkos:
@geerlingguy does it have to do maybe with [this](https://github.com/geerlingguy/ansible-role-docker/commit/c3a127134d6bde44bdd57fa1be1751f77d887614#diff-23e21b6c89468f7534697ad091835b3ee5e0213b37ace489488234e5a9548ec4R41) commit? was the initial name of the list file `download_docker_com_linux_ubuntu.list`

kawadeomkar:
Hey @pantelis-karamolegkos I had the same issue and fixed it here https://github.com/geerlingguy/ansible-role-docker/pull/436 
do you mind trying it out to see if it works for you?
You first need to run 
[...]

PR #436, opened 2023-11-17
Add GPG keys to `/apt/keyrings` instead of `/apt/trusted.gpg.d`, Update task `Add Docker apt key`

kawadeomkar:
https://github.com/geerlingguy/ansible-role-docker/issues/435

I started this issue because I faced a similar problem as https://github.com/geerlingguy/ansible-role-docker/issues/434
when adding the docker apt repository
`Signed-By regarding source https://download.docker.com/linux/ubuntu/ jammy:
│ /etc/apt/trusted.gpg.d/docker.asc != ,`
Which I then had to fix by:
```
sudo rm /etc/apt/sources.list.d/docker.list
sudo rm /etc/apt/sources.list.d/download_docker_com_linux_ubuntu.list
```
and then running the changes in this branch.

I did some research and noticed that GPG keys should not be placed in `/etc/apt/trusted.gpg.d/` (please see https://github.com/geerlingguy/ansible-role-docker/issues/435)

After updating the tasks to mirror the installation method in https://docs.docker.com/engine/install/debian/#install-using-the-repository, I was still getting an error:
```
W: GPG error: https://download.docker.com/linux/ubuntu jammy InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 7EA0A9C3F273FCD8
E: The repository 'https://download.docker.com/linux/ubuntu jammy InRelease' is not signed.
```
until I noticed that the task here https://github.com/geerlingguy/ansible-role-docker/blob/8ff4a241477f5f24c718d606a2ee450b370cc47c/tasks/setup-Debian.yml#L30
was saving the GPG key in binary whereas in the docker docs, the key is converted to an ASCII-encoded format.

I updated the ansible tasks to use the fallback curl/shell method and added a few more tasks to mirror the installation method of the docker docs.

With these changes, I'm not getting any errors and able to successfully run the role on version 7.0.2.

@geerlingguy (or any maintainer) Please let me know what you think of these changes:
1. whether the GPG keys should be saved to `/etc/apt/keyrings` (as per docker) or `/usr/share/keyrings` (as per debian)
2. If the tasks proposed in this PR are sufficient
If using the task `ansible.builtin.get_url` is preferred, I believe a task like this might be required as well directly afterwards:
```
- name: Dearmor Docker GPG key
  ansible.builtin.command: gpg --dearmor /etc/apt/keyrings/docker.gpg
```

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -33,13 +33,16 @@
     state: directory
     mode: '0755'
 
-- name: Ensure curl is present
-  package: name=curl state=present
-
-- name: Add Docker apt key
-  shell: >
-    curl -fsSL {{ docker_apt_gpg_key }} | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg --yes
-  changed_when: false
+- name: Add Docker apt key.
+  ansible.builtin.get_url:
+    url: "{{ docker_apt_gpg_key }}"
+    dest: /etc/apt/keyrings/docker.asc
+    mode: '0644'
+    force: false
+    checksum: "{{ docker_apt_gpg_key_checksum | default(omit) }}"
+  register: add_repository_key
+  ignore_errors: "{{ docker_apt_ignore_key_error }}"
+  when: docker_add_repo | bool
 
 - name: Change permissions for /etc/apt/keyrings/docker.gpg
   file:
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -39,6 +39,7 @@
 - name: Add Docker apt key 
   shell: >
     curl -fsSL {{ docker_apt_gpg_key }} | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg --yes
+  changed_when: false
 
 - name: Change permissions for /etc/apt/keyrings/docker.gpg 
   file:
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -44,10 +44,14 @@
   ignore_errors: "{{ docker_apt_ignore_key_error }}"
   when: docker_add_repo | bool
 
-- name: Change permissions for /etc/apt/keyrings/docker.gpg
-  file:
-    path: /etc/apt/keyrings/docker.gpg
-    mode: 'a+r'
+- name: Ensure curl is present (on older systems without SNI).
+  package: name=curl state=present
+  when: add_repository_key is failed and docker_add_repo | bool
+
+- name: Add Docker apt key (alternative for older systems without SNI).
+  shell: >
+    curl -sSL {{ docker_apt_gpg_key }} | apt-key add -
+  when: add_repository_key is failed and docker_add_repo | bool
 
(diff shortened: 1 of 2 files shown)


bryion:
This worked for me!

tasks/setup-Debian.yml
```diff
@@ -27,25 +27,24 @@
...
   package: name=curl state=present
-  when: add_repository_key is failed and docker_add_repo | bool
 
-- name: Add Docker apt key (alternative for older systems without SNI).
+- name: Add Docker apt key
   shell: >
-    curl -sSL {{ docker_apt_gpg_key }} | apt-key add -
-  when: add_repository_key is failed and docker_add_repo | bool
+    curl -fsSL {{ docker_apt_gpg_key }} | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg --yes
+  changed_when: false
```
kawadeomkar, line 42: not sure of a way around this

jantari:
This also affects Ubuntu.

@geerlingguy sorry for the ping but this role is currently not only not working, every time it is used and rolled out it re-breaks the docker repositories and `apt update` on existing machines.

tasks/setup-Debian.yml
jantari: Yes, there is: https://stackoverflow.com/questions/71585303/how-can-i-manage-keyring-files-in-trusted-gpg-d-with-ansible-playbook-since-apt

You can just skip the `gpg --dearmor` and save the file directly into `/etc/apt/keyrings/` with a `.asc` extension. This means you can skip the `shell` module completely and only use `get_url`.

jantari: Really all that needs to be changed here is the `dest: /etc/apt/trusted.gpg.d/docker.asc` from the `Add Docker apt key.` task needs to be `dest: /etc/apt/keyrings/docker.asc` instead.

geerlingguy:
See: https://www.jeffgeerling.com/blog/2022/aptkey-deprecated-debianubuntu-how-fix-ansible as well...

kawadeomkar:
Thank you @jantari and @geerlingguy for the information. I will update this PR and @ you again when it is ready.

kawadeomkar:
Hello, apologies for the delay. I was out on vacation with limited internet connectivity. I've updated the PR to only contain edits to save to `/apt/keyrings` using the original `get_url` ansible command. @jantari Please take a look whenever you get the chance, thanks!

rizlas:
This also fix raspbian compatibility. 
For raspbian it is also need to change these variables:

```
docker_apt_arch: armhf
docker_apt_ansible_distribution: raspbian
```

https://docs.docker.com/engine/install/raspberry-pi-os/#install-using-the-repository

kawadeomkar:
@geerlingguy sorry for the tag but could you take a look whenever you get a chance? Thanks

mihalt:
> @geerlingguy sorry for the tag but could you take a look whenever you get a chance? Thanks

it would be pritty wonderful. I've got the same problem.

jamesmyatt:
LGTM

PR #439, opened 2023-12-28
Fix docker compose plugin on archlinux (Closes #427)

### LINKED ISSUE #427: docker-compose-plugin install fails on arch
As the title says, the `Install docker-compose plugin` task fails on Arch Linux (not sure for how long, but about a week ago it still worked. The Error message is `error: 'docker-compose-plugin': could not find or read package`, so I guess that the package was renamed or replaced, although I couldn't find any more information on that.

For now, I temporarily solved this by removing the task manually from the file, but that is of course not a permanent solution.

cschindlbeck:


Javex:
Any chance this could be merged? I just ran into the same issue. It's easy enough to work around it, but would be great if the role understood it natively 😄

cschindlbeck:
Can you take a look here @geerlingguy ?

PR #440, opened 2024-01-04
Update README.md due to docker-py issue

Jan-Titze:
https://github.com/docker/docker-py/issues/3194/ forces downgrade to docker 6.1.3.

Readme update only. Maybe link to the issue in the readme as well, but i thought that was overkill.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Many users of this role wish to also use Ansible to then _build_ Docker images a
 
   vars:
     pip_install_packages:
-      - name: docker
+      - name: "docker < 7.0.0"
 
   roles:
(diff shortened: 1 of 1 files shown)


geerlingguy:
Is this related to docker_compose not working? If so, see https://github.com/ansible-collections/community.docker/pull/739 — Ansible's main docker collection is going to move away from the python library: https://github.com/ansible-collections/community.docker/pull/739

Jan-Titze:
Yes, I also think this is related to the docker_compose v1 deprecation.
Seems like the workaround is only needed until community.docker 3.6.0 release.

PR #466, opened 2024-08-25
README: Updated docker compose and docker compose plugin default values

Yemtex:
README now shows, that this role is using the recommended docker compose plugin


geerlingguy:
Indeed! Thank you :)

PR #469, opened 2024-09-22
Update README.md

lanedif:
Saw some errors in my install that devicemapper was deprecated and that overlay2 is the recommended default: https://docs.docker.com/engine/storage/drivers/select-storage-driver/


geerlingguy:
Thanks!

PR #473, opened 2024-11-03
Update variables to support Kali Linux

straysheep-dev:
After seeing how Pop!_OS and Linux Mint were supported, it looks like the same can be done for Kali using a separate variable for the release string here:

```yml
docker_apt_ansible_distribution_release: "{{ 'bookworm' if ansible_distribution in ['Kali'] else ansible_distribution_release }}"
```

...and replacing the few instances of `ansible_distribution_release` with that new variable. 

There may be a better way to do this since `bookworm` will need updated with each [Debian major release](https://www.debian.org/releases/). However it's infrequent enough that it may be acceptable for now, similar to the GPG key checksum being a static value.


clem9669:
Hey @straysheep-dev  @geerlingguy  👋 

This looks great ! I hope this will be merge soon 🙏 

Cheers

mr-pmillz:
When will this PR be merged? @clem9669 
I'm getting failing builds for installing docker on Kali via this role due to the wrong url getting set for this task: https://github.com/geerlingguy/ansible-role-docker/blob/master/tasks/setup-Debian.yml#L40

Issue #499, opened 2025-03-12
Even further backwards compatibility issues with download_docker_com_linux_debian.list


alexander-haller:
On my system running Debian12 I have this file leftover.

`/etc/apt/sources.list.d/download_docker_com_linux_debian.list`

The system is part of a GitLab installation and the Issue was first raised here: 
https://gitlab.com/gitlab-org/gitlab-environment-toolkit/-/issues/1046#note_2388089541

My current assumption is that the `docker_apt_ansible_distribution` variable is missing a `| lower ` filter and that is why the lowercase debian never gets deleted.

See also:
https://gitlab.com/gitlab-org/gitlab-environment-toolkit/-/issues/1046#note_2389115628

> docker_apt_ansible_distribution this seems to me is missing a | lower filter in the task
> It is defined in defaults as such:
> docker_apt_ansible_distribution: "{{ 'ubuntu' if ansible_distribution in ['Pop!_OS', 'Linux Mint'] else ansible_distribution }}"
> ansible_distribution should be "Debian" the file is /etc/apt/sources.list.d/download_docker_com_linux_debian.list (with a lowercase "debian")

geerlingguy:
Fixed via PR above.

PR #500, opened 2025-03-19
fix: lowercase distro name when removing apt source

boite:
This makes removal succeed when docker_apt_ansible_distribution is
"Ubuntu".


boite:
Fixes #499 also.

PR #507, opened 2025-05-30
add docker service configuration options and custom trust private Docker registries

afreisinger:
Custom docker service configuration. Should only be used for `HTTP/HTTPS proxy` settings.


afreisinger:
custom trust private Docker registries is included

PR #511, opened 2025-07-21
Refactor Debian setup to use deb822_repository

sebdanielsson:
- Replaces deprecated apt_repository and manual GPG key management with ansible.builtin.deb822_repository, introduced in apt 1.1.
- Updates handlers to use FQCNs.

Should be compatible with all supported Ubuntu and Debian distros. deb822 was introduced in apt 1.1.

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -51,8 +51,3 @@
 
 - name: Ensure handlers are notified immediately to update the apt cache.
   ansible.builtin.meta: flush_handlers
-
-- name: Remove Docker apt key if docker_add_repo is false.
-  ansible.builtin.file:
-    path: /etc/apt/keyrings/docker.asc
-    state: "{{ 'file' if docker_add_repo | bool else 'absent' }}"
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -49,6 +49,9 @@
     state: "{{ 'present' if docker_add_repo | bool else 'absent' }}"
   notify: apt update
 
+- name: Ensure handlers are notified immediately to update the apt cache.
+  ansible.builtin.meta: flush_handlers
+
 - name: Remove Docker apt key if docker_add_repo is false.
   ansible.builtin.file:
     path: /etc/apt/keyrings/docker.asc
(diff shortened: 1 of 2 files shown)


tasks/setup-Debian.yml
```diff
@@ -10,57 +10,49 @@
     state: absent
 
 - name: Ensure the repo referencing the previous trusted.gpg.d key is not present
-  apt_repository:
+  ansible.builtin.apt_repository:
     repo: "deb [arch={{ docker_apt_arch }} signed-by=/etc/apt/trusted.gpg.d/docker.asc] {{ docker_repo_url }}/{{ docker_apt_ansible_distribution | lower }} {{ ansible_distribution_release }} {{ docker_apt_release_channel }}"
     state: absent
     filename: "{{ docker_apt_filename }}"
     update_cache: true
   when: docker_add_repo | bool
```
acieslewicz, line 18: Should this be removed/combined with the previous step? It should be possible to simply delete "/etc/apt/sources.list.d/{{ docker_apt_filename }}.list" since its replaced with the deb822 file.

tasks/setup-Debian.yml
```diff
@@ -10,57 +10,49 @@
...
+
 - name: Ensure dependencies are installed.
-  apt:
+  ansible.builtin.apt:
     name:
       - apt-transport-https
       - ca-certificates
+      - curl
+      - gpg
+      - python3-debian
```
acieslewicz, line 38: I think with the removed steps and if the apt_repository check is remove, curl and gpg are no longer needed. Is python3 a dependency that needs to be checked here?

tasks/setup-Debian.yml
```diff
@@ -10,57 +10,49 @@
...
+    state: "{{ 'present' if docker_add_repo | bool else 'absent' }}"
+  notify: apt update
+
+- name: Ensure handlers are notified immediately to update the apt cache.
+  ansible.builtin.meta: flush_handlers
+
+- name: Remove Docker apt key if docker_add_repo is false.
+  ansible.builtin.file:
+    path: /etc/apt/keyrings/docker.asc
+    state: "{{ 'file' if docker_add_repo | bool else 'absent' }}"
```
acieslewicz, line 58: This is handled by the deb822_repository builtin even if only the key exists, so I don't think this task is needed.

sebdanielsson:
Yes! That's correct. I'll create a PR asap. Also can confirm that the role works great on Debian Trixie so might as well add that to the list☺️

sebdanielsson:
@gizero @geerlingguy PR: #514

PR #516, opened 2025-09-03
Specify RedHat-specific repo

### LINKED ISSUE #515: Yum repos must be different for CentOS and RHEL
According to the Docker maintainers (see https://github.com/moby/moby/issues/49169#issuecomment-3247827127), starting with RHEL8, there are now separate repos for RHEL and CentOS. As a result, these lines don't set the correct repo URL and GPG key URL:

https://github.com/geerlingguy/ansible-role-docker/blob/c0ddaa9d646e7dc7e10152559446c7d6ce378136/defaults/main.yml#L59
https://github.com/geerlingguy/ansible-role-docker/blob/c0ddaa9d646e7dc7e10152559446c7d6ce378136/defaults/main.yml#L62

I'll try to submit a PR shortly.

LoicGombeaud:
Fixes #515


LoicGombeaud:
Checks failed, but it seems to be an unrelated error (happening after the molecule "destroy" stage):

```
INFO     default ➜ destroy: Completed
INFO     default ➜ scenario: Pruning extra files from scenario ephemeral directory
Error: Process completed with exit code 2.
```

PR #522, opened 2025-10-15
Allow the start command of the Docker service to be overwritten.

chrisdecker1201:
I needed to set the 'docker_daemon_options' to the following:

```yml
docker_daemon_options : {
      "hosts": ["unix:///var/run/docker.sock", "tcp://0.0.0.0:2375"]
    }
```

Afterwards, I could no longer start the Docker service, because the default ExecStart-Command on Rocky Linux 9 is `ExecStart=/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock`.

This results in the following error message:

```txt
unable to configure the Docker daemon with file /etc/docker/daemon.json: the following directives are specified both as a flag and in the configuration file: hosts: (from flag: [fd://], from file: [unix:///var/run/docker.sock tcp://0.0.0.0:2375])
```

I tried to fix this issue with the pull request :smile: . Thank you so much ❤️.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -68,6 +68,25 @@
   when: docker_daemon_options.keys() | length > 0
   notify: restart docker
 
+- name: Replace Docker service ExecStart command.
+  when: docker_service_start_command != ""
+  notify: restart docker
+  block:
+    - name: Get Docker service status
+      ansible.builtin.systemd_service:
+        name: docker
+      register: docker_service_status
+    - name: Patch docker.service
+      ansible.builtin.replace:
+        path: "{{ docker_service_status.status['FragmentPath'] }}"
+        regexp: "^ExecStart=.*$"
+        replace: "ExecStart={{ docker_service_start_command }}"
+      register: docker_service_patch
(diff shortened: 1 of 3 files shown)


tasks/main.yml
```diff
@@ -68,11 +68,26 @@
...
+        path: "{{ docker_service_status.status['FragmentPath'] }}"
+        regexp: "^ExecStart=.*$"
+        replace: "ExecStart={{ docker_service_start_command }}"
+
 - name: Ensure Docker is started and enabled at boot.
   service:
     name: docker
     state: "{{ docker_service_state }}"
     enabled: "{{ docker_service_enabled }}"
+    daemon_reload: true
```
geerlingguy: Would this cause the daemon to be reloaded on every run?

chrisdecker1201: Yes. I just looked into the documentation. I will look for a better solution :smile: 

---
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/systemd_service_module.html#parameter-daemon_reload

PR #535, opened 2026-01-20
Moving the creation of /etc/docker and /etc/docker/daemon.json

tomn46037:
Moving the creation of /etc/docker and /etc/docker/daemon.json to before the install.  The default-address-pools option must be in place for the install or docker will create the network with the wrong IP network.  This is because the networks are created during the installation of docker.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -19,6 +19,21 @@
 - include_tasks: setup-Debian.yml
   when: ansible_facts.os_family == 'Debian'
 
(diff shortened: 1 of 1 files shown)


tomn46037:
Hi, kinda new to git/github and pull requests.  :) I thought I was merging this into my master branch.  But this was the eventual goal anyway.

If this isn't the right way to do this, or you need any changes - I'm happy to update it.

Thanks for all the work you put in!

PR #541, opened 2026-04-13
Add docker-compose plugin support for openSUSE/SLES

floh96:
## Summary

On SLES/openSUSE, the Compose v2 CLI plugin is distributed as the \`docker-compose\` package (via openSUSE OBS \`Virtualization:/containers\`), not \`docker-compose-plugin\` as on Debian/Ubuntu systems. The package installs the binary to \`/usr/lib/docker/cli-plugins/\`, making \`docker compose\` available as a Docker CLI subcommand.

Previously, \`Suse\` was explicitly excluded from both compose plugin installation tasks in \`tasks/main.yml\`, so \`docker_install_compose_plugin: true\` had no effect on SLES/openSUSE.

## Changes

- \`vars/Suse.yml\`: sets \`docker_compose_package: docker-compose\` to override the Debian/Ubuntu default of \`docker-compose-plugin\`
- \`tasks/setup-Suse.yml\`: adds a dedicated compose plugin install task using \`ansible.legacy.zypper\` with \`changed_when: false\`
- \`tasks/main.yml\`: keeps \`Suse\` in the exclusion list — the generic \`package\` module with zypper incorrectly reports \`changed\` even when the package is already installed, breaking idempotency; the dedicated task in \`setup-Suse.yml\` handles this instead

## Testing

Verified on SLES 15 SP7 — \`docker compose version\` returns \`Docker Compose version 2.22.0\` after installation. Idempotent on re-run (no spurious \`changed\`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/vars/Suse.yml
+++ b/vars/Suse.yml
@@ -37,5 +37,10 @@ docker_suse_release: >-
 docker_zypper_repo_url: >-
   https://download.opensuse.org/repositories/Virtualization:/containers/{{ docker_suse_release | trim }}/
 
+# On SLES/openSUSE, the compose v2 CLI plugin is packaged as 'docker-compose'
+# (installed to /usr/lib/docker/cli-plugins/). This differs from the Debian/Ubuntu
+# package name 'docker-compose-plugin' used by Docker Inc.'s own repositories.
+docker_compose_package: docker-compose
+
 # Control whether to add the Docker repository
 docker_add_repo: true
--- a/tasks/setup-Suse.yml
+++ b/tasks/setup-Suse.yml
@@ -23,6 +23,18 @@
   register: zypper_refresh
   changed_when: false  # idempotent for Molecule
 
+# Install docker-compose plugin (openSUSE / SLES)
+# Uses ansible.legacy.zypper instead of the generic package module to ensure
+# idempotent change reporting — zypper may report changed even when the package
+# is already installed when called via the generic module.
+- name: Install docker-compose plugin.
+  ansible.legacy.zypper:
+    name: "{{ docker_compose_package }}"
+    state: "{{ docker_compose_package_state }}"
+  notify: restart docker
+  when: docker_install_compose_plugin | bool
+  changed_when: false  # idempotent for Molecule
+
 # Install Docker packages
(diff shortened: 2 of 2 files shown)


floh96:
@geerlingguy Could you please review? 
I tested the change on SLES 15 SP7 server.

floh96:
@geerlingguy could you please review the pr?

vars/Suse.yml
```diff
@@ -37,5 +37,10 @@ docker_suse_release: >-
 docker_zypper_repo_url: >-
   https://download.opensuse.org/repositories/Virtualization:/containers/{{ docker_suse_release | trim }}/
 
+# On SLES/openSUSE, the compose v2 CLI plugin is packaged as 'docker-compose'
+# (installed to /usr/lib/docker/cli-plugins/). This differs from the Debian/Ubuntu
+# package name 'docker-compose-plugin' used by Docker Inc.'s own repositories.
+docker_compose_package: docker-compose
```
geerlingguy, line 43: Is this universal to all SLES installs? Or something specific to the base install?

floh96: @geerlingguy  Good question — no, it's not universal in the way it looks in this PR alone.

This role uses two different paths to get Docker onto SLES/openSUSE, and the availability of `docker-compose` differs between them. I've tested both on SLES 15 SP7:

1. **This PR's default path** (`docker_add_repo: true`, `docker_suse_use_containers_module: false`): Docker itself comes from the openSUSE OBS community repo (`Virtualization:containers`). That same repo also ships `docker-compose`, so it "just works" here.
2. **The official path added in #542** (`docker_suse_use_containers_module: true`): Docker/containerd/runc come from SUSE's officially supported `sle-module-containers` via SCC — but that module does **not** include `docker-compose`. It's only available via **SUSE Package Hub**, activated via an extra `SUSEConnect --product PackageHub/...` step before installing.

So to be precise: neither path gives you a fully "officially supported" source for `docker-compose` — it's either the OBS repo (this PR) or SUSE Package Hub (#542). `docker_compose_package: docker-compose` is the correct package name in both cases, just from a different repo depending on the install path.

PR #544, opened 2026-04-21
Allow overriding Debian Docker APT repo URL

jeleel-muibi:
Adds docker_apt_repo_url and uses it in the Debian deb822_repository task so internal mirrors or caches can override the APT repo URL without replacing the full repository definition. Fixes #537.


jeleel-muibi:
Thanks, Jeff. Glad this was useful.

PR #546, opened 2026-09-04
Add optional rootless Docker support

je4yfc:
Adds opt-in rootless Docker support while leaving the existing rootful path unchanged by default.

When `docker_rootless: true`, the role:

* installs the OS-specific rootless prerequisites;
* requires a non-root `docker_rootless_user`, (if it does not already exist)
* derives the user's UID, GID, home directory, runtime directory, and Docker socket.
* enables the kernel user-namespace settings required by Docker rootless mode.
* preserves existing `/etc/subuid` and `/etc/subgid` allocations and creates an allocation only when one does not already exist;
* enables systemd linger for the daemon user;
* manages `~/.config/docker/daemon.json` using the existing `docker_daemon_options` interface;
* renders and manages Docker directly as a systemd user service using `dockerd-rootless.sh`;
* stops, disables, and masks the system `docker.service` and `docker.socket`;
* exposes the effective rootless daemon endpoint as `docker_rootless_socket`.

When rootless mode is disabled, the normal system Docker units are unmasked before the existing rootful service-management path runs, allowing hosts to move back to rootful Docker cleanly.

### Notes: 
- RootlessKit uses Docker’s normal network-driver selection unless docker_rootless_network_driver is set, in which case the role installs any required package for the selected driver. Additional daemon environment variables can be passed through docker_rootless_environment. 

- Existing daemon users are reused as-is where possible, with their home directory reconciled for rootless Docker, while existing subordinate UID/GID ranges are preserved and only missing allocations are created.

- Privileged-port support uses community.general.capabilities to grant cap_net_bind_service; if you want to avoid that collection dependency, @geerlingguy, I can switch it.

- A dedicated rootless scenario molocule covers the default RootlessKit path, explicit `slirp4netns` provisioning, daemon/socket behavior, rootless Docker operation, and conflicting network-driver configuration.

- Rootless mode is currently supported on systemd-based Debian, RedHat, and SUSE families.

### Old variables: 

The existing service controls continue to apply to whichever Docker daemon mode is selected:

* `docker_service_manage`
* `docker_service_state`
* `docker_service_enabled`
* `docker_restart_handler_state`
* `docker_daemon_options`

### New variables

```yaml
docker_rootless: false
docker_rootless_user: ""
docker_rootless_subid_start: ""
docker_rootless_subid_count: 65536
docker_rootless_network_driver: ""
docker_rootless_environment: {}
docker_rootless_expose_privileged_ports: false
```


je4yfc:
Moved it to a draft while I solve the molecule issues. It seems to be issues with certain services not being started, which is understandable for a container, but unless we want the docker role to support rootless installation inside a molecule - which seems a touch OTT to me - are mostly issues with the tests and not the role if I'm understanding things correctly. 

I'm just having a case of Friday Night Works-On-My-Machine: 

| Distribution | Scenario | Converge | Idempotence | Verify | 
| :--- | :--- | :--- | :--- | :--- | 
| `ubuntu2404` | `rootless` | Passed | Passed | Passed (71/71 assertions ok) | 
| `fedora43` | `rootless` | Passed | Passed | Passed (72/72 assertions ok) | 
| `opensuseleap15` | `rootless` | Passed | Passed | Passed (67/67 assertions ok) | 

Edit: Okay, fixed. Tested against a fedora VM, Ubuntu VM, and a Debian VM. Seems to work fine. Hopefully not another case of WOMM. 
Edit 2: Of course, didn't test opensuse. FML. 😂

